PR #94, opened 2026-04-11
feat(roxabi-sdk): #93 extract browser bootstrap to roxabi_sdk.browser

### LINKED ISSUE #93: Extract browser bootstrap (playwright + stealth) to roxabi_sdk/browser.py
## Problem

Two plugins currently duplicate the `playwright` + `playwright-stealth` bootstrap:

| Plugin | File | Variant | Context lifecycle | Purpose |
|---|---|---|---|---|
| `web-intel` | `scripts/fetchers/stealth.py` | **sync** | ephemeral | anti-bot fallback for `/scrape` |
| `web-intel` | `scripts/screenshot.py` | **sync** | ephemeral | full-page PNG for `/roast` and `/benchmark` |
| `linkedin-apply` | `scripts/scraper.py:265-314` (`get_browser_context`) | **async** | `launch_persistent_context` | LinkedIn job scraping with session reuse |

Overlap per variant: ~30-40 LoC (import guards, stealth instantiation, UA/viewport config, webdriver-off args).

Divergence: sync vs async, ephemeral vs persistent, different stealth flag combinations.

Bug fixes already apply to one side and not the other — e.g. the error-surfacing refactor in commits 80387a7..7f58dbd is not mirrored in linkedin-apply.

## Proposed solution

New module `roxabi_sdk/browser.py` with two entry points:

```python
def launch_stealth_sync(
    *,
    user_data_dir: str | None = None,   # None → ephemeral, path → persistent
    headless: bool = True,
    viewport: dict | None = None,
    stealth_config: dict | None = None,
) -> tuple[Playwright, BrowserContext, Page]: ...

async def launch_stealth_async(...) -> tuple[Playwright, BrowserContext, Page]: ...
```

- Raises a typed `PlaywrightNotAvailableError` with install instructions on `ImportError`
- Applies `playwright-stealth` patches if available, with optional override flags
- Returns the tuple so callers own the lifecycle (close/cleanup)
- `user_data_dir=None` → ephemeral `new_context`; `user_data_dir=path` → `launch_persistent_context`

## Refactor scope

1. `roxabi_sdk/browser.py` — new module (sync + async variants)
2. `plugins/web-intel/scripts/fetchers/stealth.py` — replace inline bootstrap with `launch_stealth_sync()`
3. `plugins/web-intel/scripts/screenshot.py` — same
4. `plugins/linkedin-apply/scripts/scraper.py` — replace `get_browser_context()` with `launch_stealth_async(user_data_dir=LINKEDIN_PROFILE_DIR, ...)`
5. Unit tests for both variants (mocked `ImportError`, real launch smoke test)
6. Doctor updates in web-intel

## Risks

- **linkedin-apply's LinkedIn session handling is auth-sensitive.** Refactor must preserve cookie/profile semantics. Live LinkedIn test required.
- **web-intel's stealth module is freshly shipped.** Behavior change must preserve the `(html, error)` tuple return contract from `fetch_html_stealth()`.

## Context

Discovered while auditing web-intel's stealth fallback in commits:
- `80387a7` feat(web-intel): add stealth browser fallback
- `f7ada4a` refactor(web-intel): drop dormant agent-browser path
- `51a0995` feat(web-intel): add Playwright screenshot fallback for roast/benchmark
- `7f58dbd` fix(web-intel): surface stealth failures + trigger retry on 5xx anti-bot

Decision deferred to its own task for scope discipline — see session thread for the full analysis.

MickaelV0:
## Summary
- **New shared module** `roxabi_sdk/browser.py` deduplicates the playwright + playwright-stealth bootstrap that previously lived in three call sites (web-intel × 2 sync, linkedin-apply × 1 async). Exports `launch_stealth_{sync,async}`, `close_stealth{,_async}`, and the typed `PlaywrightNotAvailableError(RuntimeError)`.
- **Three latent bugs fixed** in the same diff (see V0 probe below): linkedin-apply's never-worked `use_async(...).start()` AttributeError path, and web-intel's silently-no-op `use_sync(page)` stealth in both `stealth.py` and `screenshot.py`.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #93: Extract browser bootstrap (playwright + stealth) to roxabi_sdk/browser.py | OPEN |
| Frame | [93-extract-browser-bootstrap-frame.mdx](artifacts/frames/93-extract-browser-bootstrap-frame.mdx) | Approved (F-full) |
| Analysis | [93-extract-browser-bootstrap-analysis.mdx](artifacts/analyses/93-extract-browser-bootstrap-analysis.mdx) | Present |
| Spec | [93-extract-browser-bootstrap-spec.mdx](artifacts/specs/93-extract-browser-bootstrap-spec.mdx) | Approved |
| Plan | [93-extract-browser-bootstrap-plan.mdx](artifacts/plans/93-extract-browser-bootstrap-plan.mdx) | Approved (26 micro-tasks) |
| Implementation | 3 commits on `feat/93-extract-browser-bootstrap` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ⚠️ (see notes) | See "Test Plan" |

## Behavior-equivalence probe

_V0 probe — Settled Decision #4 validation (SC-19)_

Before any code was written, a behavior-equivalence probe ran against `playwright-stealth==2.0.2` to confirm that `Stealth().apply_stealth_sync(context)` (the spec's chosen idiom) actually patches the navigator surface. Output:

```
======================================================================
Revised V0 probe — #93 browser-bootstrap extraction
======================================================================

[GATE] apply_stealth_sync(context) — spec idiom (Settled Decision #4):
   navigator.webdriver            = False
   navigator.plugins.length       = 3
   navigator.languages            = ['en-US', 'en']
   typeof window.chrome           = 'object'

[INFO] use_sync(page) post-creation — current web-intel pattern:
   navigator.webdriver            = True
   navigator.plugins.length       = 0
   navigator.languages            = ['en-US']
   typeof window.chrome           = 'undefined'
   → NOT patched (placebo). Failures: 4

======================================================================
PASS — apply_stealth_sync(context) produces stealth-patched surface.
Settled Decision #4 is validated. T0.1 is GREEN.
======================================================================
```

The `[INFO]` block is the crucial side-finding: `Stealth().use_sync(page)` called **after** `context.new_page()` is a **no-op** in `playwright-stealth==2.0.2`. That is exactly the pattern web-intel was using on lines 168 (stealth.py) and 107 (screenshot.py). So the existing web-intel anti-bot fallback has been running effectively unpatched. Both call sites are now actually stealth-patched via the SDK launcher, which calls `apply_stealth_sync(context)` **before** `new_page()`.

## Latent bugs fixed

1. **`linkedin-apply/scripts/scraper.py:306`** — `await stealth.use_async(async_playwright()).start()` raises `AttributeError: 'AsyncWrappingContextManager' object has no attribute 'start'` in `playwright-stealth==2.0.2`. This code path has **never worked** since the v2 upgrade. Now uses `launch_stealth_async(user_data_dir=…)`.
2. **`web-intel/scripts/fetchers/stealth.py:168`** — `_Stealth().use_sync(page)` after `context.new_page()` is a no-op (probe-verified). Now applied to context before page creation.
3. **`web-intel/scripts/screenshot.py:107`** — same pattern, same fix.

## Architecture (target → ref)

```
roxabi_sdk/browser.py
  ├─ launch_stealth_sync(user_data_dir, headless, ua, viewport, locale, stealth_flags, launch_args)  → (pw, ctx, page)
  ├─ launch_stealth_async(...)                                                                       → (pw, ctx, page)
  ├─ close_stealth(pw, ctx)         — picks browser.close vs context.close based on shape
  ├─ close_stealth_async(pw, ctx)
  ├─ PlaywrightNotAvailableError(RuntimeError)
  └─ _raise_if_unavailable()        — eager probe used by both launchers

plugins/web-intel/scripts/fetchers/stealth.py     → launch_stealth_sync + close_stealth
plugins/web-intel/scripts/screenshot.py            → launch_stealth_sync + close_stealth
plugins/linkedin-apply/scripts/scraper.py          → launch_stealth_async + alias class
```

Both web-intel files keep their `PLAYWRIGHT_AVAILABLE` / `PLAYWRIGHT_STEALTH_AVAILABLE` module flags **independent** of the SDK probe so the existing `tests/test_stealth.py:104` monkey-patch keeps working unchanged (SC-5).

The linkedin-apply alias is multiple-inheritance:

```python
class PlaywrightNotAvailableError(LinkedInScraperError, _SdkPlaywrightNotAvailableError):
    ...
```

Left-first MRO → `__init__` resolves to `LinkedInScraperError(message, url=None)`, so `self.url` will be `None` for SDK-raised instances. Documented in the class docstring.

## Test Plan

- [x] `roxabi_sdk/browser.py` — 7 mock-based tests in `tests/test_browser.py` (no live Playwright). Covers missing-dep typed error, ephemeral vs persistent close in sync + async, finally-clause invariant, default launch kwargs.
- [x] `plugins/web-intel/tests/test_stealth.py` — 23 passed, file untouched (T2.3 SC-5).
- [x] `plugins/web-intel/tests/test_screenshot.py` — 6 passed.
- [x] **CLI smoke**: `python scripts/screenshot.py https://example.com /tmp/shot_93.png` → 20 KB PNG written.
- [x] `plugins/linkedin-apply/tests/test_alias.py` — 3 new tests: catchable as `LinkedInScraperError`, catchable as SDK error, `self.url is None`. RED before T4.2, all green after.
- [x] **Grep assertion**: `grep -c 'use_async\|.start()' plugins/linkedin-apply/scripts/scraper.py` → `0` (SC-11).
- [x] **Non-bootstrap grep counts unchanged** (SC-14): web-intel/stealth.py 8↔8, linkedin-apply/scraper.py 21↔21.
- [x] **Import guard** (SC-17): `from roxabi_sdk.browser import PlaywrightNotAvailableError` works in a venv with playwright + playwright_stealth blocked from `sys.meta_path`. Typed error raises with install hint.
- [x] **Packaging** (SC-18): `rsync -avn roxabi_sdk/ ~/.claude/.../web-intel/0.1.0/roxabi_sdk/` shows `browser.py` would land via the existing line 71-72 of `sync-plugins.sh` (no script changes — SC-16: `git diff staging -- sync-plugins.sh` → 0).
- [ ] **LinkedIn manual smoke (T4.5)** — `cd plugins/linkedin-apply && python scripts/scraper.py <real-job-url>` against an authenticated profile. Confirm JSON has non-null `job_id`/`title`/`company`. Paste full JSON below before merging.

### LinkedIn smoke

> ⏸ Pending — needs human attestation. See checklist item above.

## Notes for the reviewer

- **bun test on this branch: 329 pass / 9 fail / 2 errors.** All failures are in `plugins/dev-core` TS suites (`issues` GraphQL batching, `scaffoldFumadocs`, `config`). Branch diff is **Python-only** — zero TS files touched. Failures exist on `staging` baseline, not introduced by this PR. Verified with `git diff staging --stat` (6 files, all `.py`).
- **CI runs `python3 -m pytest tests/`** at the repo root. `tests/test_browser.py` is included and green. `plugins/web-intel/tests/` and `plugins/linkedin-apply/tests/` are NOT run by CI today — they were verified locally only (gap noted, not in scope for #93).
- **Plan drift**: the `## Task IDs` section in `artifacts/plans/93-...` was reattached this session (commit `95f73dd` on staging) because the previous task ids had been reclaimed by the fresh `/dev` pipeline. The actual implementation is unchanged.
- **Plan T1.4 snippet had a missing `await`** on `apply_stealth_async(ctx)` — the function is a coroutine in 2.0.2 (`inspect.iscoroutinefunction == True`). Fixed in the actual code; plan doc nit only.

Closes #93

---

### DIFF
--- a/plugins/linkedin-apply/scripts/scraper.py
+++ b/plugins/linkedin-apply/scripts/scraper.py
@@ -187,6 +187,7 @@ class ScraperTimeoutError(LinkedInScraperError):
 # launcher will carry self.url=None, which is harmless but non-obvious.
 from roxabi_sdk.browser import (  # noqa: E402
     PlaywrightNotAvailableError as _SdkPlaywrightNotAvailableError,
+    launch_stealth_async as _sdk_launch_stealth_async,
 )
 
 
@@ -295,17 +296,12 @@ async def get_browser_context(
         PlaywrightNotAvailableError: If playwright or playwright-stealth is
             not installed (catchable as :class:`LinkedInScraperError` too).
     """
-    from roxabi_sdk.browser import (
-        PlaywrightNotAvailableError as _SdkPWError,
-        launch_stealth_async,
-    )
-
     try:
-        playwright, context, _page = await launch_stealth_async(
+        playwright, context, _page = await _sdk_launch_stealth_async(
             user_data_dir=LINKEDIN_PROFILE_DIR,
             headless=headless,
         )
-    except _SdkPWError as exc:
+    except _SdkPlaywrightNotAvailableError as exc:
         raise PlaywrightNotAvailableError(str(exc)) from exc
 
     logger.info("Browser context created with stealth mode")
--- a/plugins/linkedin-apply/scripts/scraper.py
+++ b/plugins/linkedin-apply/scripts/scraper.py
@@ -34,7 +34,7 @@ logger = logging.getLogger(__name__)
 
 # Type checking imports
 if TYPE_CHECKING:
-    from playwright.async_api import Page, BrowserContext
+    from playwright.async_api import BrowserContext, Page, Playwright as AsyncPlaywright
 
 
 # ============================================================================
@@ -187,6 +187,7 @@ class ScraperTimeoutError(LinkedInScraperError):
 # launcher will carry self.url=None, which is harmless but non-obvious.
 from roxabi_sdk.browser import (  # noqa: E402
     PlaywrightNotAvailableError as _SdkPlaywrightNotAvailableError,
+    close_stealth_async as _sdk_close_stealth_async,
     launch_stealth_async as _sdk_launch_stealth_async,
 )
 
@@ -284,11 +285,11 @@ def validate_linkedin_url(url: str) -> bool:
 
 async def get_browser_context(
     headless: bool = False,
-) -> tuple[Any, "BrowserContext"]:
+) -> tuple[AsyncPlaywright, BrowserContext, Page]:
     """Create a stealth-patched persistent LinkedIn browser context.
 
     Delegates to :func:`roxabi_sdk.browser.launch_stealth_async` with the
-    LinkedIn profile directory. Returns ``(playwright, context)`` — the
+    LinkedIn profile directory. Returns ``(playwright, context, page)`` — the
     caller owns the lifecycle and must call
     :func:`roxabi_sdk.browser.close_stealth_async` when done.
 
@@ -297,7 +298,7 @@ async def get_browser_context(
             not installed (catchable as :class:`LinkedInScraperError` too).
     """
     try:
-        playwright, context, _page = await _sdk_launch_stealth_async(
+        playwright, context, page = await _sdk_launch_stealth_async(
             user_data_dir=LINKEDIN_PROFILE_DIR,
             headless=headless,
         )
@@ -305,7 +306,7 @@ async def get_browser_context(
         raise PlaywrightNotAvailableError(str(exc)) from exc
 
     logger.info("Browser context created with stealth mode")
-    return playwright, context
+    return playwright, context, page
 
 
 async def check_login(page: "Page") -> bool:
@@ -398,12 +399,8 @@ async def scrape_job(
     context = None
 
     try:
-        # Get browser context
-        playwright, context = await get_browser_context(headless=headless)
-
-        # Get or create page
-        pages = context.pages
-        page = pages[0] if pages else await context.new_page()
+        # Get browser context + stealth-patched page from SDK
+        playwright, context, page = await get_browser_context(headless=headless)
 
         # Navigate to job page
         logger.debug("Navigating to job page...")
@@ -526,17 +523,11 @@ async def scrape_job(
         return job
 
     finally:
-        # Cleanup
-        if context:
+        if playwright and context:
             try:
-                await context.close()
+                await _sdk_close_stealth_async(playwright, context)
             except Exception as e:
-                logger.warning(f"Error closing context: {e}")
-        if playwright:
-            try:
-                await playwright.stop()
-            except Exception as e:
-                logger.warning(f"Error stopping playwright: {e}")
+                logger.warning(f"Error during browser cleanup: {e}")
 
 
 async def scrape_job_with_retry(
--- a/plugins/linkedin-apply/scripts/scraper.py
+++ b/plugins/linkedin-apply/scripts/scraper.py
@@ -181,10 +181,29 @@ class ScraperTimeoutError(LinkedInScraperError):
     pass
 
 
-class PlaywrightNotAvailableError(LinkedInScraperError):
-    """Raised when Playwright is not installed."""
+# Import the SDK error so the local alias can be caught from both
+# hierarchies (see #93 spec S3a). MRO is left-first, so __init__ resolves
+# to LinkedInScraperError(message, url=None) — instances raised by the SDK
+# launcher will carry self.url=None, which is harmless but non-obvious.
+from roxabi_sdk.browser import (  # noqa: E402
+    PlaywrightNotAvailableError as _SdkPlaywrightNotAvailableError,
+)
 
-    pass
+
+class PlaywrightNotAvailableError(
+    LinkedInScraperError, _SdkPlaywrightNotAvailableError
+):
+    """Alias error catchable from both LinkedIn and SDK exception hierarchies.
+
+    Allows ``except PlaywrightNotAvailableError`` and
+    ``except LinkedInScraperError`` clauses to both catch the same instance,
+    so existing LinkedIn callers keep working while cross-cutting SDK
+    callers (e.g. dashboards that aggregate browser-bootstrap failures)
+    can also handle it uniformly with web-intel.
+
+    ``self.url`` will be ``None`` for errors raised by the SDK launcher
+    (the SDK does not know which URL the caller was about to fetch).
+    """
 
 
 # ============================================================================
@@ -265,57 +284,29 @@ def validate_linkedin_url(url: str) -> bool:
 async def get_browser_context(
     headless: bool = False,
 ) -> tuple[Any, "BrowserContext"]:
-    """
-    Create browser context with persistent profile and stealth mode.
+    """Create a stealth-patched persistent LinkedIn browser context.
 
-    Returns:
-        Tuple of (playwright instance, browser context)
+    Delegates to :func:`roxabi_sdk.browser.launch_stealth_async` with the
+    LinkedIn profile directory. Returns ``(playwright, context)`` — the
+    caller owns the lifecycle and must call
+    :func:`roxabi_sdk.browser.close_stealth_async` when done.
 
     Raises:
-        PlaywrightNotAvailableError: If playwright or playwright-stealth not installed
-
-    TODO(#93): The Playwright + stealth bootstrap below duplicates the sync
-    variant in ``plugins/web-intel/scripts/fetchers/stealth.py`` and
-    ``plugins/web-intel/scripts/screenshot.py``. Extract to
-    ``roxabi_sdk/browser.py`` as ``launch_stealth_async(user_data_dir=...)``
-    so all three sites share one primitive. See Roxabi/roxabi-plugins#93
-    for the full refactor plan.
+        PlaywrightNotAvailableError: If playwright or playwright-stealth is
+            not installed (catchable as :class:`LinkedInScraperError` too).
     """
+    from roxabi_sdk.browser import (
+        PlaywrightNotAvailableError as _SdkPWError,
+        launch_stealth_async,
+    )
+
     try:
-        from playwright.async_api import async_playwright
-        from playwright_stealth import Stealth
-    except ImportError as e:
-        raise PlaywrightNotAvailableError(
-            f"Playwright or playwright-stealth not installed: {e}\n"
-            "Run: pip install playwright playwright-stealth && playwright install chromium"
+        playwright, context, _page = await launch_stealth_async(
+            user_data_dir=LINKEDIN_PROFILE_DIR,
+            headless=headless,
         )
-
-    # Ensure profile directory exists
-    os.makedirs(LINKEDIN_PROFILE_DIR, exist_ok=True)
-
-    # Create stealth instance
-    stealth = Stealth(
-        navigator_webdriver=True,
-        chrome_runtime=True,
-        navigator_plugins=True,
-        navigator_permissions=True,
-        webgl_vendor=True,
-    )
-
-    # Use stealth-instrumented playwright
-    playwright = await stealth.use_async(async_playwright()).start()
-
-    # Launch persistent context
-    context = await playwright.chromium.launch_persistent_context(
-        user_data_dir=LINKEDIN_PROFILE_DIR,
-        headless=headless,
-        viewport={"width": 1280, "height": 900},
-        args=[
-            "--disable-blink-features=AutomationControlled",
-            "--no-first-run",
-            "--no-default-browser-check",
-        ],
-    )
+    except _SdkPWError as exc:
+        raise PlaywrightNotAvailableError(str(exc)) from exc
 
     logger.info("Browser context created with stealth mode")
     return playwright, context
--- a/plugins/linkedin-apply/tests/test_alias.py
+++ b/plugins/linkedin-apply/tests/test_alias.py
@@ -52,3 +52,9 @@ def test_alias_has_url_attribute_none() -> None:
     """
     exc = PlaywrightNotAvailableError("test msg")
     assert exc.url is None
+
+
+def test_alias_url_kwarg_stored() -> None:
+    """When a ``url`` is provided, the LinkedIn-side ``__init__`` must store it."""
+    exc = PlaywrightNotAvailableError("test msg", url="https://linkedin.com/jobs/123")
+    assert exc.url == "https://linkedin.com/jobs/123"
--- a/plugins/linkedin-apply/tests/test_alias.py
+++ b/plugins/linkedin-apply/tests/test_alias.py
@@ -1,0 +1,54 @@
+"""Alias catchability tests for ``PlaywrightNotAvailableError``.
+
+After #93, ``plugins.linkedin-apply.scripts.scraper.PlaywrightNotAvailableError``
+must be catchable as both:
+
+* :class:`scripts.scraper.LinkedInScraperError` — for callers that
+  ``except LinkedInScraperError`` to handle every scraper failure uniformly.
+* :class:`roxabi_sdk.browser.PlaywrightNotAvailableError` — for callers that
+  ``except PlaywrightNotAvailableError`` to handle the missing-dep case
+  cross-cutting all browser users.
+
+The pre-refactor class only inherited from ``LinkedInScraperError``, so the
+SDK-side ``isinstance`` check below is the RED phase: it must fail before
+T4.2 introduces the multiple-inheritance alias and pass after.
+"""
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+
+# Match the sys.path bootstrapping used by tests/test_adapters.py.
+_plugin_root = str(Path(__file__).resolve().parents[1])
+_repo_root = str(Path(__file__).resolve().parents[3])
+for _p in (_plugin_root, _repo_root):
+    if _p not in sys.path:
+        sys.path.insert(0, _p)
+
+from roxabi_sdk.browser import PlaywrightNotAvailableError as SdkPWError
+from scripts.scraper import (
+    LinkedInScraperError,
+    PlaywrightNotAvailableError,
+)
+
+
+def test_alias_is_catchable_as_linkedin_error() -> None:
+    """The alias must still satisfy ``except LinkedInScraperError``."""
+    exc = PlaywrightNotAvailableError("test msg")
+    assert isinstance(exc, LinkedInScraperError)
+
+
+def test_alias_is_catchable_as_sdk_error() -> None:
+    """The alias must also satisfy ``except SdkPWError`` after T4.2."""
+    exc = PlaywrightNotAvailableError("test msg")
+    assert isinstance(exc, SdkPWError)
+
+
+def test_alias_has_url_attribute_none() -> None:
+    """MRO is left-first → ``LinkedInScraperError.__init__(message, url=None)``.
+
+    The alias does not pass a ``url`` so ``self.url`` should be ``None``.
+    Documented in plan T4.2: harmless but non-obvious for SDK errors.
+    """
+    exc = PlaywrightNotAvailableError("test msg")
+    assert exc.url is None
--- a/plugins/web-intel/scripts/fetchers/stealth.py
+++ b/plugins/web-intel/scripts/fetchers/stealth.py
@@ -8,12 +8,6 @@ fallback utility invoked by ``generic.py`` when the fast path (plain HTTP
 via ``safe_fetch`` + Trafilatura) fails to extract meaningful content or
 hits an anti-bot signature.
 
-TODO(#93): The Playwright + stealth bootstrap below (launch → context →
-page → apply stealth patches) duplicates the async variant in
-``plugins/linkedin-apply/scripts/scraper.py:get_browser_context``. Extract
-to ``roxabi_sdk/browser.py`` as ``launch_stealth_sync()`` so both plugins
-share one primitive. See Roxabi/roxabi-plugins#93 for the full refactor plan.
-
 Triggers for a stealth retry:
   - HTTP 403 / 429 / 503 from the fast path
   - Cloudflare / generic anti-bot challenge markers in the body
@@ -35,7 +29,6 @@ from __future__ import annotations
 import logging
 import sys
 from pathlib import Path
-from typing import Optional, Tuple
 
 # Add _shared to path for sibling imports (consistent with other fetchers)
 SHARED_DIR = Path(__file__).resolve().parents[1] / "_shared"
@@ -97,9 +90,9 @@ except ImportError:
 
 
 def has_antibot_signature(
-    status_code: Optional[int] = None,
-    html: Optional[str] = None,
-    text_length: Optional[int] = None,
+    status_code: int | None = None,
+    html: str | None = None,
+    text_length: int | None = None,
 ) -> bool:
     """Detect whether a fast-path fetch looks like it hit anti-bot protection.
 
@@ -126,7 +119,7 @@ def has_antibot_signature(
 def fetch_html_stealth(
     url: str,
     timeout_ms: int = DEFAULT_TIMEOUT_MS,
-) -> Tuple[Optional[str], Optional[str]]:
+) -> tuple[str | None, str | None]:
     """Fetch a URL's rendered HTML using headless Chromium + stealth patches.
 
     Intended as a fallback for anti-bot protected pages. Returns a
--- a/plugins/web-intel/scripts/fetchers/stealth.py
+++ b/plugins/web-intel/scripts/fetchers/stealth.py
@@ -77,15 +77,27 @@ _DEFAULT_USER_AGENT = (
 )
 _DEFAULT_VIEWPORT = {"width": 1280, "height": 900}
 
+from roxabi_sdk.browser import (
+    PlaywrightNotAvailableError,
+    close_stealth,
+    launch_stealth_sync,
+)
+
+# These two probe flags stay independent of the SDK so the existing
+# test suite (test_stealth.py:104) can keep monkey-patching
+# PLAYWRIGHT_AVAILABLE to drive the "missing dep" branch without having
+# to reach into roxabi_sdk internals. They serve a different role from
+# roxabi_sdk.browser._raise_if_unavailable: this is a module-level
+# pre-flight gate, not the eager import probe used by the launcher.
 try:
-    from playwright.sync_api import sync_playwright
+    import playwright  # noqa: F401
 
     PLAYWRIGHT_AVAILABLE = True
 except ImportError:
     PLAYWRIGHT_AVAILABLE = False
 
 try:
-    from playwright_stealth import Stealth as _Stealth
+    import playwright_stealth  # noqa: F401
 
     PLAYWRIGHT_STEALTH_AVAILABLE = True
 except ImportError:
@@ -153,47 +165,45 @@ def fetch_html_stealth(
         logger.warning("Stealth fetch refused: %s", msg)
         return None, msg
 
+    pw = ctx = None
     try:
-        with sync_playwright() as p:
-            browser = p.chromium.launch(headless=True)
-            try:
-                context = browser.new_context(
-                    user_agent=_DEFAULT_USER_AGENT,
-                    viewport=_DEFAULT_VIEWPORT,
-                    locale="en-US",
-                )
-                page = context.new_page()
+        try:
+            pw, ctx, page = launch_stealth_sync(
+                user_agent=_DEFAULT_USER_AGENT,
+                viewport=_DEFAULT_VIEWPORT,
+                locale="en-US",
+            )
+        except PlaywrightNotAvailableError as exc:
+            # SDK probe failed even though our PLAYWRIGHT_AVAILABLE flag passed
+            # — surface the SDK install hint verbatim so the user can act on it.
+            logger.info("Stealth fallback unavailable: %s", exc)
+            return None, str(exc)
 
-                if PLAYWRIGHT_STEALTH_AVAILABLE:
-                    _Stealth().use_sync(page)
-                    logger.debug("playwright-stealth patches applied")
-                else:
-                    logger.info(
-                        "playwright-stealth not installed; stealth fetch running "
-                        "without fingerprint patches (less effective)"
-                    )
+        page.goto(url, timeout=timeout_ms, wait_until="domcontentloaded")
+        # Let dynamic content settle — CF challenge pages typically
+        # auto-redirect after ~1-2s when the stealth patches work
+        page.wait_for_timeout(POST_LOAD_WAIT_MS)
 
-                page.goto(url, timeout=timeout_ms, wait_until="domcontentloaded")
-                # Let dynamic content settle — CF challenge pages typically
-                # auto-redirect after ~1-2s when the stealth patches work
-                page.wait_for_timeout(POST_LOAD_WAIT_MS)
+        html = page.content()
 
-                html = page.content()
-            finally:
-                browser.close()
+        # If we STILL see a challenge marker, stealth didn't bypass it —
+        # don't pretend success, let the caller surface a clear error
+        for marker in CF_CHALLENGE_MARKERS:
+            if marker in html:
+                msg = f"still blocked by anti-bot challenge after stealth retry ({marker!r})"
+                logger.info("Stealth fetch: %s", msg)
+                return None, msg
 
-            # If we STILL see a challenge marker, stealth didn't bypass it —
-            # don't pretend success, let the caller surface a clear error
-            for marker in CF_CHALLENGE_MARKERS:
-                if marker in html:
-                    msg = f"still blocked by anti-bot challenge after stealth retry ({marker!r})"
-                    logger.info("Stealth fetch: %s", msg)
-                    return None, msg
-
-            return html, None
+        return html, None
 
     except Exception as exc:
         # Preserve exception type so callers can see e.g. "TimeoutError: ..."
         msg = f"{type(exc).__name__}: {exc}"
         logger.warning("Stealth fetch failed for %s: %s", url, msg)
         return None, msg
+    finally:
+        if pw is not None and ctx is not None:
+            try:
+                close_stealth(pw, ctx)
+            except Exception:
+                logger.debug("close_stealth raised during cleanup", exc_info=True)
--- a/plugins/web-intel/scripts/fetchers/stealth.py
+++ b/plugins/web-intel/scripts/fetchers/stealth.py
@@ -69,14 +69,6 @@ DEFAULT_TIMEOUT_MS = 30_000
 # Wait after domcontentloaded for dynamic content / CF auto-redirect
 POST_LOAD_WAIT_MS = 2_500
 
-# Viewport + UA tuned to match a common desktop Chrome fingerprint
-_DEFAULT_USER_AGENT = (
-    "Mozilla/5.0 (Windows NT 10.0; Win64; x64) "
-    "AppleWebKit/537.36 (KHTML, like Gecko) "
-    "Chrome/120.0.0.0 Safari/537.36"
-)
-_DEFAULT_VIEWPORT = {"width": 1280, "height": 900}
-
 from roxabi_sdk.browser import (
     PlaywrightNotAvailableError,
     close_stealth,
@@ -168,15 +160,11 @@ def fetch_html_stealth(
     pw = ctx = None
     try:
         try:
-            pw, ctx, page = launch_stealth_sync(
-                user_agent=_DEFAULT_USER_AGENT,
-                viewport=_DEFAULT_VIEWPORT,
-                locale="en-US",
-            )
+            pw, ctx, page = launch_stealth_sync()
         except PlaywrightNotAvailableError as exc:
             # SDK probe failed even though our PLAYWRIGHT_AVAILABLE flag passed
             # — surface the SDK install hint verbatim so the user can act on it.
-            logger.info("Stealth fallback unavailable: %s", exc)
+            logger.info("Stealth fallback unavailable: %s", exc, exc_info=True)
             return None, str(exc)
 
         page.goto(url, timeout=timeout_ms, wait_until="domcontentloaded")
--- a/plugins/web-intel/scripts/screenshot.py
+++ b/plugins/web-intel/scripts/screenshot.py
@@ -28,9 +28,12 @@ Security:
 
 from __future__ import annotations
 
+import logging
 import sys
 from pathlib import Path
 from typing import Tuple
+
+logger = logging.getLogger(__name__)
 
 # Add _shared to path for sibling imports (consistent with fetchers/*.py)
 SHARED_DIR = Path(__file__).resolve().parent / "_shared"
@@ -44,13 +47,6 @@ DEFAULT_TIMEOUT_MS = 30_000
 
 # Wait after domcontentloaded for dynamic content / fonts to settle
 POST_LOAD_WAIT_MS = 2_500
-
-_DEFAULT_USER_AGENT = (
-    "Mozilla/5.0 (Windows NT 10.0; Win64; x64) "
-    "AppleWebKit/537.36 (KHTML, like Gecko) "
-    "Chrome/120.0.0.0 Safari/537.36"
-)
-_DEFAULT_VIEWPORT = {"width": 1280, "height": 900}
 
 from roxabi_sdk.browser import (  # noqa: E402
     PlaywrightNotAvailableError,
@@ -106,11 +102,7 @@ def capture_full_page(
     pw = ctx = None
     try:
         try:
-            pw, ctx, page = launch_stealth_sync(
-                user_agent=_DEFAULT_USER_AGENT,
-                viewport=_DEFAULT_VIEWPORT,
-                locale="en-US",
-            )
+            pw, ctx, page = launch_stealth_sync()
         except PlaywrightNotAvailableError as exc:
             return False, str(exc)
 
@@ -127,7 +119,7 @@ def capture_full_page(
             try:
                 close_stealth(pw, ctx)
             except Exception:
-                pass
+                logger.debug("close_stealth raised during screenshot cleanup", exc_info=True)
 
 
 def main() -> int:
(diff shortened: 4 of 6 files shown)


MickaelV0:
## Code Review — PR #94 (#93 extract browser bootstrap)

5 fresh review agents reviewed this PR independently. Findings consolidated, deduped (same `file:line + label` → max confidence kept), grouped by category, sorted by confidence within group. Verdict at the bottom.

**Stats:** 5 agents · 38 findings · 10 blockers · 11 warnings · 8 thoughts/questions · 5 praise

---

### Blockers — must address before merge

#### `issue(blocking):` SC-13 — LinkedIn smoke JSON absent from PR; section is "Pending"
- **PR #94 body — `### LinkedIn smoke` section**
- `-- product-lead`
- **Root cause:** Spec SC-13 is unambiguous: full JSON dict with non-null `job_id`, `title`, `company` must be pasted before merging. The "Pending — needs human attestation" placeholder does not satisfy this requirement. Slice 4 verification is incomplete.
- **Solutions:**
  1. Run `cd plugins/linkedin-apply && python scripts/scraper.py <real-job-url>` against an authenticated profile, paste full JSON into the PR description (recommended)
  2. Defer to a follow-up PR — requires spec revision
- **Confidence: 100%**

#### `issue:` `test_launch_kwargs_defaults_applied` line 208 has a tautology assertion that always passes
- `tests/test_browser.py:208`
- `-- tester`
- **Root cause:** `assert (pw, ctx, page) == (pw, ctx, page)` compares a tuple to itself. The intent was probably to assert the returned values are the fakes injected by `_install_fake_playwright`. As written this assertion can never fail and provides zero signal — T6 currently passes by accident.
- **Solutions:**
  1. Capture the fakes from `_install_fake_playwright` (return them) and assert `pw is fake_pw and ctx is fake_context and page is fake_page` (recommended)
  2. Remove the line; rest of the test verifies mock calls which implicitly confirm identity
- **Confidence: 99%**

#### `issue:` `launch_stealth_async` has zero unit tests
- `tests/test_browser.py` (absent)
- `-- tester`
- **Root cause:** Only `close_stealth_async` received tests. Because `_install_fake_playwright` builds a sync chain only, there is no async-API fake. The `await Stealth(...).apply_stealth_async(ctx)` call is exactly the kind of place the V0 probe found a no-op bug — not testing it means the new API can regress to that same no-op silently.
- **Solutions:**
  1. Add `_install_fake_playwright_async(monkeypatch)` mirroring the sync helper but with `AsyncMock` for `async_playwright().start()`, `launch`, `new_context`, `apply_stealth_async`. Assert `apply_stealth_async.assert_awaited_once_with(ctx)` (recommended)
  2. Use `unittest.mock.patch` with `AsyncMock` directly inside the test for a lighter setup
- **Confidence: 96%**

#### `issue:` `close_stealth` untyped parameters allow caller to pass arguments in reversed order
- `roxabi_sdk/browser.py:209,232`
- `-- backend-dev`
- **Root cause:** `playwright` and `context` carry no type annotations, so static checkers treat them as `Any`. The most common mistake when wrapping paired teardown APIs is calling `close_stealth(ctx, pw)` instead of `close_stealth(pw, ctx)`. The TYPE_CHECKING block already imports `SyncPlaywright`, `SyncBrowserContext`, `AsyncPlaywright`, `AsyncBrowserContext`.
- **Solutions:**
  1. Add `playwright: SyncPlaywright, context: SyncBrowserContext` (and async mirror) using the existing TYPE_CHECKING imports (recommended)
- **Confidence: 92%**

#### `issue:` `close_stealth` ephemeral close-raises path is uncovered
- `tests/test_browser.py` (absent — contrast `test_close_stealth_persistent_still_stops_when_close_raises` at line 76)
- `-- tester`
- **Root cause:** The bonus test exercises `ctx.close()` throwing on the persistent branch. The ephemeral branch calls `context.browser.close()` instead — if that raises, the same `finally: playwright.stop()` must run. No test covers this; a refactor that accidentally moves `browser.close()` outside the `try` would go undetected.
- **Solutions:**
  1. Add `test_close_stealth_ephemeral_still_stops_when_browser_close_raises` with `ctx.browser.close.side_effect = RuntimeError("boom")` and `pw.stop.assert_called_once()` (recommended)
  2. Parameterize the existing bonus test
- **Confidence: 92%**

#### `suggestion(blocking):` `screenshot.py` close-stealth swallowed silently — no log on cleanup failure (sibling-file inconsistency)
- `plugins/web-intel/scripts/screenshot.py:128-130`
- `-- backend-dev`
- **Root cause:** The `finally` block in `capture_full_page` catches all `close_stealth` exceptions with `except Exception: pass`. Sibling file `stealth.py` does the same teardown with `logger.debug(..., exc_info=True)`. A suppressed close failure means a leaked Chromium subprocess that only becomes visible via `ps aux` or OOM.
- **Solutions:**
  1. Mirror `stealth.py`: `logger.debug("close_stealth raised during screenshot cleanup", exc_info=True)` (recommended)
- **Confidence: 91%**

#### `issue:` `ctx.pages[0] if ctx.pages else ctx.new_page()` silently reuses an existing page in the persistent path
- `roxabi_sdk/browser.py:151,202`
- `-- backend-dev`
- **Root cause:** For a persistent profile (`user_data_dir` set), Playwright often opens the profile's previous tabs on launch, so `ctx.pages` is non-empty. The launcher silently returns the first pre-existing page rather than a fresh one. The caller (e.g. `scraper.py:get_browser_context`) immediately navigates that page, mutating a tab the user may have left open in their LinkedIn session — and means the page was created BEFORE `apply_stealth_sync` was called, defeating the documented contract. For ephemeral, `ctx.pages` is always `[]`.
- **Solutions:**
  1. Always call `ctx.new_page()` unconditionally (recommended)
  2. Keep the reuse logic but have the caller explicitly close stale pages
- **Confidence: 87%**

#### `issue:` Two independent "is playwright available?" probes with no shared source of truth
- `plugins/web-intel/scripts/fetchers/stealth.py:92-104`
- `plugins/web-intel/scripts/screenshot.py:65-77`
- `-- architect`
- **Root cause:** The module-level `PLAYWRIGHT_AVAILABLE` flags do a bare `import playwright` probe, while the SDK's `_raise_if_unavailable` does the same probe again. When playwright IS installed, both probes succeed and the module-level guard is dead weight. When NOT installed, `fetch_html_stealth` returns a different install hint string than `PlaywrightNotAvailableError.message` — DRY violation with a user-facing surface.
- **Solutions:**
  1. Expose `browser.is_available() -> bool` in the SDK (no import side effects) and replace both module-level try/except blocks (recommended)
  2. Accept the duplication but unify the message strings — alias the early-return message to `browser._INSTALL_HINT`
- **Confidence: 82%**

#### `suggestion(blocking):` SC-9 — `get_browser_context` body is 13 non-blank lines, spec requires ≤ 10
- `plugins/linkedin-apply/scripts/scraper.py:284-313`
- `-- product-lead`
- **Root cause:** The current body is 13 non-blank lines because it includes a 4-line local import block. Hoisting the SDK import to module level (where `_SdkPlaywrightNotAvailableError` is already imported at line 189) would remove the inner block and bring the body to 9 non-blank lines.
- **Solutions:**
  1. Hoist `from roxabi_sdk.browser import launch_stealth_async` to module level next to `_SdkPlaywrightNotAvailableError`. Use the already-imported `_SdkPlaywrightNotAvailableError` in the except clause (recommended)
- **Confidence: 78%**

#### `issue:` Summary — 17/19 SC met; 1 hard block (SC-13), 1 soft gap (SC-9)
- PR #94
- `-- product-lead`
- **BLOCKING:** SC-13 — LinkedIn smoke JSON not pasted. **NON-BLOCKING:** SC-9 — body size 13 vs ≤10. **STALE SPEC:** SC-5 — file says "11 tests" but actual count is 23 (file is byte-identical to staging; all green).
- **Confidence: 100%**

---

### Warnings — non-blocking suggestions and nitpicks

#### `nitpick:` Unused `call` import in `tests/test_browser.py:12`
- `-- tester` · **Root cause:** `from unittest.mock import AsyncMock, MagicMock, call` — `call` never used. Leftover from a draft.
- **Solutions:** 1. Remove `call` from the import (recommended)
- **Confidence: 99%**

#### `nitpick:` `get_browser_context` re-imports `PlaywrightNotAvailableError` inside function body
- `plugins/linkedin-apply/scripts/scraper.py:298-301`
- `-- architect` · **Root cause:** Same root cause as SC-9. The module already imports `_SdkPlaywrightNotAvailableError` at line 189; the function-level re-import creates a third alias for the same class.
- **Solutions:** 1. Remove the function-level import. Use `_SdkPlaywrightNotAvailableError` directly (recommended)
- **Confidence: 95%**

#### `nitpick:` String-quoted return types redundant with `from __future__ import annotations`
- `roxabi_sdk/browser.py:105,167`
- `-- backend-dev` · **Root cause:** `from __future__ import annotations` causes all annotations to be treated as strings at runtime. Explicitly quoting `"tuple[SyncPlaywright, SyncBrowserContext, SyncPage]"` is double-string and signals uncertainty about how `__future__` annotations work.
- **Solutions:** 1. Remove the outer quotes — write `-> tuple[SyncPlaywright, SyncBrowserContext, SyncPage]` unquoted (recommended)
- **Confidence: 95%**

#### `suggestion(non-blocking):` `import os` deferred inside function body — inconsistent style
- `roxabi_sdk/browser.py:136,187`
- `-- architect` · **Root cause:** `os` is stdlib; deferring its import inside the `if user_data_dir` branch adds no startup benefit. A reader may wonder if there is a deliberate lazy-import rationale (there isn't — unlike playwright, `os` carries no optional-dependency concern).
- **Solutions:** 1. Move `import os` to the module-level top section (recommended)
- **Confidence: 92%**

#### `nitpick:` `_DEFAULT_USER_AGENT` / `_DEFAULT_VIEWPORT` shadow-copied in both web-intel files instead of removed
- `plugins/web-intel/scripts/fetchers/stealth.py:73-78`
- `plugins/web-intel/scripts/screenshot.py:48-53`
- `-- architect` · **Root cause:** Both web-intel local constants now exist in parallel to `_DEFAULT_UA` / `_DEFAULT_VIEWPORT` in the SDK. The call sites pass the local copies as kwargs to the launcher — web-intel still "owns" the values rather than inheriting SDK defaults. Soft fork of the very data the extraction was meant to centralize.
- **Solutions:** 1. Delete the local constants from both web-intel files; let `launch_stealth_sync()` fall back to SDK defaults (pass no `user_agent`/`viewport`) (recommended)
- **Confidence: 88%**

#### `suggestion(non-blocking):` SC-15 — PR description conflates `bun test` and `bun run test`; misleading note about 9 failures
- PR #94 body — "Notes for the reviewer" section
- `-- product-lead` · **Root cause:** `bun test` (Bun's native runner) hits TS suites with pre-existing failures on staging unrelated to this PR. CI uses `bun run test` (vitest) which passes 349/349. A reviewer could interpret the 9 failures as blocking.
- **Solutions:** 1. Prepend "bun run test (vitest): 349/349 pass" before the bun-native-runner note (recommended)
- **Confidence: 85%**

#### `suggestion(non-blocking):` `stealth.py` exception chain swallowed when catching `PlaywrightNotAvailableError`
- `plugins/web-intel/scripts/fetchers/stealth.py:176-180`
- `-- backend-dev` · **Root cause:** The inner `except PlaywrightNotAvailableError as exc: return None, str(exc)` strips the original `ImportError` text chained via `from exc`. The logged message includes only the outer message, not the inner `ModuleNotFoundError: No module named 'playwright_stealth'` that tells the user exactly which dep is missing.
- **Solutions:** 1. `logger.info("Stealth fallback unavailable: %s", exc, exc_info=True)` so debug log shows the chain (recommended)
- **Confidence: 84%**

#### `suggestion(non-blocking):` `test_alias_url_kwarg_stored` missing — non-None `url` path is unverified
- `plugins/linkedin-apply/tests/test_alias.py:47-54`
- `-- tester` · **Root cause:** Current test only exercises `url=None` default through MRO; `PlaywrightNotAvailableError("msg", url="https://x.com").url == "https://x.com"` is unverified. If `LinkedInScraperError.__init__` is ever refactored to drop the `url` kwarg, the alias silently loses it with no failing test.
- **Solutions:** 1. Add `test_alias_url_kwarg_stored` (recommended)
- **Confidence: 78%**

#### `suggestion(non-blocking):` `_DEFAULT_STEALTH_FLAGS = dict(...)` inconsistent with the rest of the file
- `roxabi_sdk/browser.py:87`
- `-- backend-dev` · **Root cause:** Every other constant uses literal syntax (`_DEFAULT_VIEWPORT = {"width": ...}`); `dict(key=val)` reads as a constructor call rather than a data literal. `paths.py` (sibling SDK module) uses `{}` literals exclusively.
- **Solutions:** 1. Replace with `{"navigator_webdriver": True, ...}` (recommended)
- **Confidence: 72%**

#### `suggestion(non-blocking):` `_INSTALL_HINT` may leak server filesystem paths via wrapped `ImportError`
- `roxabi_sdk/browser.py:73`
- `-- security-auditor` · **Root cause:** `_raise_if_unavailable()` builds the message as `f"{_INSTALL_HINT} ({exc})"`. On a partial install Python's `ImportError` message can include an absolute filesystem path. This string is propagated verbatim through `stealth.py:180` → `generic.py:172` → `generic.py:204` into the JSON `error` key returned to the agent caller — surfacing the server's virtual-env layout to the LLM.
- **Solutions:** 1. Strip the inner `exc` detail: `raise PlaywrightNotAvailableError(_INSTALL_HINT) from exc` — `__cause__` still carries the path for `logging.exception` but it never lands in the error string (recommended)
- **Confidence: 72%**

#### `suggestion(non-blocking):` `output_path` flows from `sys.argv` to `page.screenshot()` with no path anchor check
- `plugins/web-intel/scripts/screenshot.py:120,143`
- `-- security-auditor` · **Root cause:** `main()` passes `sys.argv[2]` directly to `capture_full_page()`. As a local CLI the immediate attack surface is the invoking user only; if `capture_full_page` is ever called from a networked agent handler that maps user-supplied filenames, arbitrary file write under the process user is possible.
- **Solutions:** 1. Add `Path(output_path).resolve().is_relative_to(ALLOWED_BASE)` guard or enforce `.png` suffix (recommended)
- **Confidence: 62%**

---

### Thoughts and questions — design choices worth flagging

#### `thought:` `PLAYWRIGHT_STEALTH_AVAILABLE` flag is now dead code in both web-intel files
- `plugins/web-intel/scripts/fetchers/stealth.py:99-104`, `screenshot.py:72-77`
- `-- backend-dev` · **Root cause:** Pre-refactor this flag gated `_Stealth().use_sync(page)`. Post-refactor `launch_stealth_sync` owns that logic; no consumer in either file. The "preserved for monkey-patching" comment is wrong — `test_stealth.py:104` patches `PLAYWRIGHT_AVAILABLE`, not `PLAYWRIGHT_STEALTH_AVAILABLE`.
- **Solutions:** 1. Remove the flag (recommended)
- **Confidence: 78%**

#### `thought:` Caller-owned 3-tuple lifecycle leaks resource management responsibility (no context manager)
- `roxabi_sdk/browser.py:151-152`
- `-- architect` · **Root cause:** Every call site replicates the same finally-guard pattern (already copy-pasted in `stealth.py:168-209` and `screenshot.py:106-130`). Spec deliberately declined a context manager (decision #3), but the result is that the "paired teardown" contract is only enforced by documentation, not by the type system.
- **Solutions:** 1. Add a context-manager variant alongside the tuple return — `stealth_session_sync()` as a `@contextmanager` (recommended) 2. Document a usage guard
- **Confidence: 78%**

#### `thought:` SC-19 / V0 probe — a real E2E test gated on `RUN_E2E=1` would protect against future no-op regressions
- `tests/` (absent)
- `-- tester` · **Root cause:** Mock-boundary philosophy — current tests verify wiring but cannot verify that the real `Stealth.apply_stealth_sync` actually patches the context, which is the core correctness guarantee. The V0 probe was scratch and now lives only as PR description text.
- **Confidence: 85%**

#### `question:` `close_stealth` uses `context.browser is None` — leaky abstraction over a Playwright internal
- `roxabi_sdk/browser.py:209-229`
- `-- architect` · **Root cause:** The persistent/ephemeral distinction is a Playwright internal invariant. If a future version changes the return value, `close_stealth` silently double-closes. The invariant is not stated in Playwright's public API contract.
- **Solutions:** 1. Return a typed pair from the launcher and have `close_stealth` use SDK-side bookkeeping (recommended) 2. Add a comment citing the Playwright source
- **Confidence: 74%**

#### `thought:` Multiple-inheritance alias makes the SDK error participate in the LinkedIn hierarchy
- `plugins/linkedin-apply/scripts/scraper.py:193-206`
- `-- architect` · **Root cause:** The SDK error is now entangled in a plugin-layer hierarchy it has no knowledge of. If a third plugin adds its own aliased subclass, the SDK error ends up entangled in N unrelated hierarchies.
- **Solutions:** 1. Keep current design (spec decision S3a) but record an ADR (recommended)
- **Confidence: 71%**

#### `question:` `isolated_vault` autouse fixture in `tests/conftest.py` applies to `test_browser.py` — coupling unclear
- `tests/conftest.py:12-18`
- `-- tester` · **Root cause:** If `browser.py` ever gains vault-relative path defaults, the autouse fixture silently redirects them to tmp. Today it adds `tmp_path` overhead per test for no reason.
- **Solutions:** 1. Verify `browser.py` does not import `roxabi_sdk.paths`; add a module-level comment confirming the fixture is a no-op for this file (recommended)
- **Confidence: 70%**

#### `question:` SC-5 — spec says "11 tests" but actual count is 23
- `plugins/web-intel/tests/test_stealth.py`
- `-- product-lead` · **Root cause:** Spec text predates additions to `test_stealth.py`. File is byte-identical to staging; all tests pass. Count discrepancy is in the spec, not the implementation.
- **Solutions:** 1. Update spec SC-5 text to current count in a follow-up cleanup commit (recommended)
- **Confidence: 92%**

#### `thought:` `self.url=None` on SDK-raised `PlaywrightNotAvailableError` is documented but unguarded
- `plugins/linkedin-apply/scripts/scraper.py:22-35`
- `-- security-auditor` · **Root cause:** No caller currently reads `exc.url` for security-sensitive purposes, but a future caller doing `logger.info("blocked URL: %s", exc.url)` would silently log `None` or raise `TypeError`. Latent risk.
- **Confidence: 60%**

---

### Praise

- `praise:` **SSRF pre-flight preserved** in both migrated callers (security-auditor, 99%) — `validate_url_ssrf(url)` runs before `launch_stealth_sync()` in both `fetch_html_stealth` and `capture_full_page`. Defense-in-depth maintained.
- `praise:` **Two-layer probe structure is sound** (architect, 95%) — `_raise_if_unavailable` + lazy body imports lets tests monkeypatch the probe independently of `sys.modules`.
- `praise:` **Exception chaining + TYPE_CHECKING pattern is well-executed** (backend-dev, 96%) — `raise … from exc` preserves the chain. `TYPE_CHECKING` ensures Playwright types are available to static checkers but never imported at runtime.
- `praise:` **`monkeypatch.setitem(sys.modules, ...)` is the right isolation primitive** (tester, 95%) — pytest restores `sys.modules` entries after each test, no cross-test leakage.
- `praise:` **Spec compliance: 17/19 SCs fully met** (product-lead, 100%) — SC-1, 2, 3, 4, 6, 7, 8, 10, 11, 12, 14, 16, 17, 19 all green.

---

## Verdict

**Request changes** — multiple blockers, including one hard product gate (SC-13).

**Most urgent (must fix before merge):**

1. **SC-13 LinkedIn smoke** — needs human action: run `scraper.py` against a real job URL with an authenticated profile, paste JSON into PR description.
2. **Tautology assertion** in `tests/test_browser.py:208` — T6 currently has zero signal; the test always passes regardless of what the launcher returned.
3. **`ctx.pages[0]` persistent reuse** in `roxabi_sdk/browser.py:151,202` — silently mutates a tab the user may have left open in their LinkedIn session AND defeats the documented "patches before any page is opened" contract.
4. **`screenshot.py` close-error swallow** — leaked Chromium subprocesses with no log signal; trivially fixable (1 line) to match the sibling `stealth.py` pattern.
5. **`launch_stealth_async` zero unit tests** — the entire async happy path is untested, including the `await Stealth(...).apply_stealth_async(ctx)` that the V0 probe specifically validated.
6. **SC-9 body size + duplicate SDK import** — fixed together by hoisting `launch_stealth_async` to module level. Drops the body to 9 lines and removes the duplicate alias.

**Quick wins (low effort, high signal):**

7. Add typed params to `close_stealth` / `close_stealth_async` (1-line change each, prevents reversed-arg bug)
8. Remove unused `call` import in `tests/test_browser.py:12`
9. Unquote return types in `roxabi_sdk/browser.py:105,167`
10. Move `import os` to module-level
11. Add ephemeral close-raises test (asymmetric coverage)
12. Delete shadow-copied `_DEFAULT_USER_AGENT`/`_DEFAULT_VIEWPORT` in both web-intel files

**Defer / discuss:**

13. Context-manager variant alongside 3-tuple return — spec decision #3 declined; revisit if a third call site appears
14. Consolidate the duplicate availability probes — needs spec amendment because the existing test monkey-patches the flag
15. ADR for the MI alias pattern in linkedin-apply — non-blocking record-keeping

---

Generated with [Claude Code](https://claude.com/claude-code) via `/code-review`

MickaelV0:
## Review Fixes Applied

**Auto-applied (lead-verified):** 17 finding(s)
**Applied via 1b1:** 0 finding(s)
**Deferred:** 2 finding(s) (SC-13 LinkedIn smoke = human gate, SC-15 PR description note = editorial)
**Skipped:** 3 (two independent probes → scope creep; output_path anchor → new feature)
**Failed:** 0 finding(s)

### Auto-Applied (17)
- [applied] `issue:` tautology assert `(pw,ctx,page) == (pw,ctx,page)` → identity checks against fakes (99%)
- [applied] `issue:` `launch_stealth_async` zero tests → added `test_launch_stealth_async_defaults_applied` with `assert_awaited_once_with` (96%)
- [applied] `issue:` `close_stealth` untyped params → added `SyncPlaywright`, `SyncBrowserContext` (and async mirror) (92%)
- [applied] `issue:` `close_stealth` ephemeral close-raises uncovered → added `test_close_stealth_ephemeral_still_stops_when_browser_close_raises` (92%)
- [applied] `suggestion(blocking):` `screenshot.py` `pass` swallows close → `logger.debug(…, exc_info=True)` + added `logging` import (91%)
- [applied] `issue:` `ctx.pages[0]` persistent reuse → always `ctx.new_page()` (87%)
- [applied] `suggestion(blocking):` SC-9 body size 13→8 lines by hoisting import to module level (78%+95%)
- [applied] `nitpick:` unused `call` import removed (99%)
- [applied] `nitpick:` string-quoted return types → unquoted (PEP 563) (95%)
- [applied] `suggestion(non-blocking):` `import os` → module-level (92%)
- [applied] `nitpick:` `_DEFAULT_USER_AGENT`/`_DEFAULT_VIEWPORT` shadow-copies deleted from both web-intel files (88%)
- [applied] `suggestion(non-blocking):` `stealth.py` exc chain → `exc_info=True` on logger (84%)
- [applied] `suggestion(non-blocking):` `test_alias_url_kwarg_stored` added (78%)
- [applied] `suggestion(non-blocking):` `_DEFAULT_STEALTH_FLAGS` dict()→{} literal (72%)
- [applied] `suggestion(non-blocking):` `_INSTALL_HINT` → stripped inner exc from message to avoid path leaks (72%)
- [applied] `nitpick:` `get_browser_context` double-import → removed function-local re-import (95%)
- [applied] `nitpick:` `stealth.py`/`screenshot.py` now call `launch_stealth_sync()` with no kwargs (SDK defaults) (88%)

### Deferred
- **SC-13:** LinkedIn smoke JSON — requires human: `cd plugins/linkedin-apply && python scripts/scraper.py <real-job-url>`. Paste output into PR description under `### LinkedIn smoke`.
- **SC-15:** PR description note about `bun test` vs `bun run test` — editorial fix on the PR body, not code.

### Skipped (out of scope for this fix iteration)
- `issue:` Two independent availability probes (82%) — affects test surface (`test_stealth.py:104` monkey-patch), needs spec amendment + careful refactoring. Deferred to follow-up.
- `suggestion(non-blocking):` `output_path` no anchor check (62%) — new feature (path validation), not a regression from #93. Deferred.
- `thought:` `PLAYWRIGHT_STEALTH_AVAILABLE` dead code — coupled to the two-probes finding above.

### Test Results Post-Fix
- `tests/test_browser.py`: 9 passed (was 7)
- `plugins/web-intel/tests/test_stealth.py` + `test_screenshot.py`: 29 passed
- `plugins/linkedin-apply/tests/test_alias.py`: 4 passed (was 3)
- `bun run lint`: ✅
- `bun run typecheck`: ✅
- `bun run test` (vitest): 349/349 ✅

MickaelV0:
## Code Review — PR #94 (#93 extract browser bootstrap) — Re-review

5 fresh review agents re-reviewed this PR after 17 fixes from round 1. Findings consolidated, deduped (same `file:line + label` → max confidence kept), grouped by category, sorted by confidence within group. Verdict at the bottom.

**Stats:** 5 agents · 16 findings · 3 blockers · 9 warnings · 0 thoughts/questions · 4 praise

---

### Blockers — must address before merge

#### `issue:` `scrape_job` teardown bypasses `close_stealth_async` — `playwright.stop()` not guaranteed on `context.close()` failure
- `plugins/linkedin-apply/scripts/scraper.py:528-539`
- `-- security-auditor, architect, backend-dev`
- **Root cause:** The `finally` block calls `context.close()` then `playwright.stop()` directly instead of delegating to `close_stealth_async(playwright, context)`. If `context.close()` raises (e.g. Playwright `TimeoutError`), `playwright.stop()` is never reached — the Playwright supervisor process leaks for the lifetime of the event loop. `close_stealth_async` wraps `playwright.stop()` in a `finally`, guaranteeing cleanup. Both `stealth.py` and `screenshot.py` already use `close_stealth` correctly.
- **Solutions:**
  1. Replace the two-step block with `await close_stealth_async(playwright, context)` inside a single `try/except` — matches the pattern in `stealth.py:192-197` and `screenshot.py:118-122` (recommended)
  2. Nest `playwright.stop()` in a `finally` inside the `context.close()` try block
- **Confidence: 88%**

#### `issue:` `scrape_job` still uses `pages[0]` fallback — may use an unpatched restored tab on persistent profiles
- `plugins/linkedin-apply/scripts/scraper.py:405-406`
- `-- backend-dev, architect`
- **Root cause:** Round 1 fixed `ctx.pages[0]` reuse inside `browser.py` (SDK always calls `ctx.new_page()`). But the call site in `scrape_job` still does `page = pages[0] if pages else await context.new_page()`. For a persistent context with restored tabs from a prior session, `context.pages[0]` picks up an old restored tab — opened BEFORE the current `apply_stealth_async(ctx)` call — defeating the documented "stealth patches applied before any page is opened" contract. The SDK-created page is at `context.pages[-1]`, not `[0]`.
- **Solutions:**
  1. Change `get_browser_context` to return `(playwright, context, page)` as a 3-tuple and have `scrape_job` use the returned page directly, matching the SDK's documented contract (recommended)
  2. Replace `pages[0]` with `await context.new_page()` unconditionally in `scrape_job`
- **Confidence: 88%**

#### `issue:` `get_browser_context` discards the SDK-returned page, creating an implicit dependency
- `plugins/linkedin-apply/scripts/scraper.py:300-308`
- `-- architect`
- **Root cause:** `launch_stealth_async` returns `(playwright, context, page)` but `get_browser_context` assigns page to `_page` and returns only `(playwright, context)`. `scrape_job` then recovers the page via `context.pages[0]` — an implicit, fragile dependency on internal Playwright state. The page created by the SDK is a guaranteed stealth-patched resource; discarding it forces the caller to guess which page is the right one.
- **Solutions:**
  1. Return `(playwright, context, page)` from `get_browser_context` and update `scrape_job` to use it directly (recommended — fixes the `pages[0]` issue above simultaneously)
  2. Document the implicit dependency explicitly
- **Confidence: 80%**

---

### Warnings — non-blocking suggestions and nitpicks

#### `issue:` Stale `TODO(#93)` comment in `stealth.py` still references extraction as future work
- `plugins/web-intel/scripts/fetchers/stealth.py:12-16`
- `-- backend-dev`
- **Root cause:** The TODO says "Extract to `roxabi_sdk/browser.py`" but the extraction is complete — file imports from `roxabi_sdk.browser` at line 72-76. Misleading to readers.
- **Solutions:** 1. Delete lines 12-16 (recommended)
- **Confidence: 97%**

#### `issue:` Identical stale `TODO(#93)` in `screenshot.py`
- `plugins/web-intel/scripts/screenshot.py:9-13`
- `-- backend-dev`
- **Root cause:** Same — extraction is done, TODO is stale.
- **Solutions:** 1. Delete lines 9-13 (recommended)
- **Confidence: 97%**

#### `suggestion(non-blocking):` Persistent launch path (`user_data_dir` set) has zero unit tests
- `tests/test_browser.py` (absent)
- `-- tester`
- **Root cause:** `_install_fake_playwright` only drives the ephemeral branch. The `if user_data_dir is not None` branch at `browser.py:136-145` and `185-194` (calls `launch_persistent_context`, `os.makedirs`) has no coverage.
- **Solutions:** 1. Add `test_launch_sync_persistent_uses_launch_persistent_context` with monkeypatched `os.makedirs` (recommended)
- **Confidence: 97%**

#### `suggestion(non-blocking):` SC-20 — Behavior-equivalence probe under wrong heading
- PR #94 body
- `-- product-lead`
- **Root cause:** Spec requires `### Behavior-equivalence probe` heading (SC-20). PR body has it under `## V0 probe — Settled Decision #4 validation (SC-19)`. Content is present, heading slug doesn't match.
- **Solutions:** 1. Rename or add alias heading in PR body (recommended — 2 min edit)
- **Confidence: 95%**

#### `issue:` `_raise_if_unavailable` actual import-failure path never exercised by tests
- `tests/test_browser.py:30-43`
- `-- tester`
- **Root cause:** The only missing-playwright test monkeypatches `_raise_if_unavailable` itself. The real body (lines 68-74 of `browser.py`) — the `try: import playwright` that catches `ImportError` and re-raises as `PlaywrightNotAvailableError` — is never executed. If the wrapping broke, all 9 tests still pass.
- **Solutions:** 1. Add a test that sets `sys.modules["playwright"] = None` and calls `_raise_if_unavailable()` directly (recommended)
- **Confidence: 92%**

#### `nitpick:` Legacy `Optional[T]` / `Tuple[…]` imports in both web-intel files
- `plugins/web-intel/scripts/fetchers/stealth.py:39`, `plugins/web-intel/scripts/screenshot.py:34`
- `-- backend-dev`
- **Root cause:** Both files have `from __future__ import annotations` — the `typing.Optional` / `typing.Tuple` imports are unnecessary. Inconsistent with `browser.py` which uses `X | None` / `tuple[A, B]` throughout.
- **Solutions:** 1. Replace with PEP 604/585 syntax and drop `typing` imports (recommended)
- **Confidence: 92%**

#### `suggestion(non-blocking):` Async close-raises paths untested
- `tests/test_browser.py` (absent)
- `-- tester`
- **Root cause:** Sync close has both persistent + ephemeral exception-propagation tests. Async `close_stealth_async` has none — if `browser.close()` or `context.close()` raises in the async path, `playwright.stop()` guarantee is unverified.
- **Solutions:** 1. Add `test_close_stealth_async_{ephemeral,persistent}_still_stops_when_close_raises` (recommended)
- **Confidence: 90%**

#### `suggestion(non-blocking):` Custom kwargs override paths untested
- `tests/test_browser.py:186-219`
- `-- tester`
- **Root cause:** Only the no-kwarg defaults path is tested. `user_agent`, `viewport`, `stealth_flags`, `launch_args` overrides never exercised — a regression where overrides are silently ignored would not be caught.
- **Solutions:** 1. Add a focused test with non-default kwargs (recommended)
- **Confidence: 85%**

#### `nitpick:` `get_browser_context` return type uses bare `Any` for playwright object
- `plugins/linkedin-apply/scripts/scraper.py:287`
- `-- backend-dev`
- **Root cause:** Now that `AsyncPlaywright` is available under `TYPE_CHECKING`, the return annotation could be `tuple[AsyncPlaywright, BrowserContext]` instead of `tuple[Any, BrowserContext]`.
- **Solutions:** 1. Add to existing `TYPE_CHECKING` block (recommended)
- **Confidence: 78%**

---

### Deferred (human gates, out of scope)

#### SC-13: LinkedIn smoke JSON — requires human attestation
- PR #94 description still has the `⏸ Pending` placeholder. Needs: run `scraper.py <real-job-url>` with authenticated profile, paste JSON.

#### Module-level `PLAYWRIGHT_AVAILABLE` flags (82%) — deferred in round 1, same rationale holds (test monkey-patching surface). Revisit when a third call site appears.

#### Sync/async launcher duplication (75%) — acceptable for initial extraction. `_build_launch_params` helper deferred to follow-up.

---

### Praise

- `praise:` **MRO for PlaywrightNotAvailableError alias is sound** (architect, 95%) — C3 linearization verified, no diamond problem. Tested by `test_alias.py`.
- `praise:` **`context.browser is None` discriminant is a stable Playwright API guarantee** (architect, 90%) — documented property, not internal detail.
- `praise:` **`_INSTALL_HINT` chain verified clean** (security-auditor, 100%) — `raise PlaywrightNotAvailableError(_INSTALL_HINT) from exc` keeps inner ImportError paths out of user-visible messages.
- `praise:` **SSRF pre-flight preserved in both web-intel callers** (security-auditor, 100%) — defense-in-depth maintained.

---

## Verdict

**Request changes** — 3 blockers, all in `plugins/linkedin-apply/scripts/scraper.py`.

**Most urgent (must fix before merge):**

1. **`scrape_job` teardown** — replace manual `context.close()` + `playwright.stop()` with `await close_stealth_async(playwright, context)`. One-liner fix, prevents resource leak.
2. **`get_browser_context` → return 3-tuple** — return `(playwright, context, page)` instead of `(playwright, context)`. Update `scrape_job` to use the returned page directly instead of `context.pages[0]`. Fixes both the discarded-page issue and the restored-tab issue simultaneously.
3. **Stale TODOs** — delete the `TODO(#93)` blocks from `stealth.py:12-16` and `screenshot.py:9-13`.

**Quick wins (low effort, high signal):**

4. SC-20 heading rename in PR body
5. Legacy `Optional`/`Tuple` → PEP 604/585 in both web-intel files

**Defer / discuss:**

6. Additional test coverage (persistent path, real import failure, async close-raises, custom kwargs) — all non-blocking, good follow-up PR
7. SC-13 LinkedIn smoke — human gate, unchanged from round 1

---

Generated with [Claude Code](https://claude.com/claude-code) via `/code-review`

MickaelV0:
## Review Fixes Applied

**Auto-applied (lead-verified):** 7 finding(s)
**Applied via 1b1:** 1 finding(s)
**Deferred:** 4 finding(s) (test coverage gaps → follow-up PR)
**Skipped:** 4 (SC-13 human gate + 3 deferred-from-round-1)
**Failed:** 0 finding(s)

### Auto-Applied (7)
- [applied] `issue:` `scrape_job` teardown bypasses `close_stealth_async` → delegate to `_sdk_close_stealth_async(playwright, context)` (88%)
- [applied] `issue:` `get_browser_context` discards SDK-returned page → return 3-tuple `(playwright, context, page)` (80%)
- [applied] `issue:` `scrape_job` `pages[0]` fallback → use returned page directly from `get_browser_context` (88%)
- [applied] `nitpick:` `get_browser_context` return type `Any` → `tuple[AsyncPlaywright, BrowserContext, Page]` with `TYPE_CHECKING` import (78%)
- [applied] `issue:` stale `TODO(#93)` in `stealth.py:12-16` → deleted (97%)
- [applied] `issue:` stale `TODO(#93)` in `screenshot.py:9-13` → deleted (97%)
- [applied] `nitpick:` legacy `Optional[T]`/`Tuple[…]` → PEP 604/585 syntax in both web-intel files (92%)

### Applied (1b1)
- [applied] `suggestion:` SC-20 probe heading → renamed to `## Behavior-equivalence probe` in PR body (95%)

### Deferred (follow-up PR)
- `suggestion:` persistent launch path untested (97%)
- `issue:` `_raise_if_unavailable` real import-failure path untested (92%)
- `suggestion:` async close-raises paths untested (90%)
- `suggestion:` custom kwargs override paths untested (85%)

### Skipped (out of scope)
- **SC-13:** LinkedIn smoke JSON — requires human: `cd plugins/linkedin-apply && uv run python scripts/scraper.py <real-job-url>`. Paste output into PR description under `### LinkedIn smoke`.
- Module-level `PLAYWRIGHT_AVAILABLE` flags (82%) — deferred in round 1, same rationale holds
- Sync/async launcher duplication (75%) — acceptable for initial extraction
- `stealth.py` `exc_info=True` at INFO level (63%) — minor log hygiene

### Test Results Post-Fix
- `tests/test_browser.py`: 9 passed
- `plugins/linkedin-apply/tests/test_alias.py`: 4 passed
- `bun run test` (vitest): 349/349 passed
- `bun run typecheck`: passed
- `trufflehog`: clean

MickaelV0:
## Code Review — PR #94 (#93 extract browser bootstrap) — Final review (iteration 3)

2 focused agents (security-auditor, product-lead) reviewed the current state after 2 fix rounds (17 + 8 findings applied).

**Stats:** 2 agents · 0 blockers · 0 warnings · 1 praise

---

### Spec Compliance — 18/19 SCs met

| SC# | Check | Status |
|-----|-------|--------|
| SC-1–8 | SDK module, error class, tests, web-intel imports, test suites, CLI smoke | met |
| SC-9 | `get_browser_context` body ≤ 10 lines | met (6 non-blank lines) |
| SC-10 | Alias unit test | met (4 tests) |
| SC-11 | `use_async`/`.start()` removed | met |
| SC-12 | Anti-detection args | met |
| SC-13 | LinkedIn smoke JSON | **deferred** (human gate) |
| SC-14–18 | Non-regression, CI, sync-plugins, import guard, packaging | met |
| SC-19 | Pre-implementation probe | met |
| SC-20 | Probe heading | met |

### Security

```
praise: No new vulnerabilities introduced by the last 2 fix commits.
  -- security-auditor
  scraper.py: 3-tuple return and close_stealth_async delegation are consistent;
  teardown guards pw/ctx before close. stealth.py + screenshot.py: sync pattern
  mirrors correctly; exceptions at debug level, no silent collapse.
  Confidence: 95%
```

---

## Verdict

**Approve** — all code blockers resolved across 3 review iterations (round 1: 10 blockers → round 2: 3 blockers → round 3: 0 blockers). 18/19 spec criteria met. SC-13 (LinkedIn smoke JSON) remains the sole human gate — not a code issue.

---

Generated with [Claude Code](https://claude.com/claude-code) via `/code-review`

Issue #10, opened 2026-03-03
Web dashboard - CI/CD Issue tracking (with boilerplate)


MickaelV0:
## Context

The issues dashboard is currently a local Bun HTTP server (localhost:3333) with SSE live updates and background polling. This issue investigates what it would take to host it as a shared team tool.

## Goal

- Shared team visibility (PM, QA, stakeholders) without per-dev local setup
- Real-time CI updates (~2s latency) via GitHub webhooks instead of polling
- No personal GitHub tokens — use a GitHub App installation token
- Optional GitHub OAuth to restrict dashboard access to org members

### Auth & Data
- **GitHub App** installed on the org — replaces `GITHUB_TOKEN` env var
  - Permissions needed: `issues:read`, `pull_requests:read`, `checks:read`, `actions:read`, `contents:read`, `projects:read`, `metadata:read`
- **Webhook handler** at `POST /webhook` (HMAC-validated) for events: `workflow_run`, `check_suite`, `pull_request`, `issues`
- **Vercel token** stays as env var (not covered by GitHub App)

### What changes vs local
| Feature | Local | Hosted |
|---|---|---|
| Auth | personal token / gh CLI | GitHub App installation token |
| CI updates | polling (60s / 20s adaptive) | webhooks (~2s) |
| Worktrees | local git | drop |
| Workspace config | file on disk | env vars |
| Access control | none (localhost) | GitHub OAuth or VPN/Tailscale |

### Deploy target
Railway or Fly.io (~$5/month or free tier on Fly). Vercel is not viable — SSE requires persistent connections.

## Out of Scope

Adaptive 20s CI polling is a separate quick-win issue to be implemented first.

## Investigation Tasks

- [ ] Confirm GitHub App can access GitHub Project V2 GraphQL data
- [ ] Map all current fetch functions to required GitHub App permissions
- [ ] Evaluate Railway vs Fly.io for this workload
- [ ] Decide on access control strategy (OAuth vs VPN)
- [ ] Estimate migration effort (polling → webhooks)

MickaelV0:
Closing — this is now tracked as the roxabi-dashboard project.

PR #100, opened 2026-04-15
fix(dev-core): resolve cwd→project via git remote and .roxabi marker

MickaelV0:
## Problem

\`roxabi issues\` fails with \`No project found for current directory\` whenever a registered repo lacks a \`localPath\` field in \`workspace.json\` — even though the repo itself *is* registered. Older versions of \`roxabi workspace add\` never wrote \`localPath\`, so most existing entries have no way to be resolved from cwd.

Concretely: \`Roxabi/roxabi-forge\` is registered in my workspace, but running \`roxabi issues\` from \`~/projects/roxabi-forge\` errors out.

## Root cause

\`resolveCurrentProject\` in \`cli/commands/issues.ts\` only matches by \`localPath\`. No fallback to git remote, \`gh\`, or any other source of truth. \`localPath\` is a *derived, machine-specific* value stored in shared config — that's the smell.

## Fix

Add \`resolveRepoFromCwd(cwd)\` with a two-step fallback (\`cli/lib/workspace.ts\`):

1. **\`.roxabi\` marker walk-up** — \`{ \"repo\": \"owner/name\" }\`. Supports monorepos where a subdir maps to a distinct project.
2. **\`git -C cwd remote get-url origin\`** — parses SSH shorthand, \`https://\`, \`ssh://\`, \`git://\`, with/without \`.git\` suffix.

\`resolveCurrentProject\` now tries \`localPath\` (exact + prefix) first, then falls back to a **case-insensitive** match on \`p.repo\` using the resolved slug. The \"not registered\" error hints at the detected repo:

\`\`\`
Detected repo: Roxabi/roxabi-forge — not registered.
Run: roxabi workspace add Roxabi/roxabi-forge
\`\`\`

Also: \`discoverProject\` now auto-populates \`localPath\` via \`detectLocalPath\` (mirroring the logic already present in \`skills/shared/adapters/workspace-helpers.ts\`), so new \`workspace add\` runs capture cwd when it matches the repo being added.

## Why this shape

- **Additive.** Existing \`localPath\` entries still match first — no behavior change for projects that already work.
- **Derives what's derivable.** Git already knows what repo you're in. Don't duplicate that in shared config.
- **Handles monorepos.** \`.roxabi\` marker is committable, travels with the code, overrides remote.
- **Works across machines.** No per-machine paths needed.

## Out of scope (deferred)

- Projects-first schema refactor (\`{ projects: [{ id, repos[], ... }] }\`) — breaking change, touches every consumer, deserves its own PR.
- Consolidating \`cli/lib/workspace.ts\` + \`skills/shared/adapters/workspace-helpers.ts\` — the two currently have overlapping duplicate logic.

## Tests

20 new tests in \`cli/__tests__/resolve.test.ts\` (all passing via \`bun test\`):

- \`parseGitRemoteUrl\` — SSH / HTTPS / with+without \`.git\` / \`ssh://\` / whitespace / invalid
- \`resolveRepoFromCwd\` — remote lookup, subdir walk-up, marker override, malformed marker fallback, non-git dir
- \`resolveCurrentProject\` — localPath exact + prefix, git-remote fallback, case-insensitive match, null case
- \`detectLocalPath\` — cwd match, \`~/projects/<name>\` scan, missing candidate

## Test plan

- [x] \`bun run typecheck\` ✅
- [x] \`bun run lint\` (biome) ✅
- [x] \`bun run test\` (vitest — doesn't cover cli/__tests__, excluded by config) ✅
- [x] \`bun test plugins/dev-core/cli/__tests__/resolve.test.ts\` → 20/20 pass
- [x] Pre-commit + pre-push hooks green
- [ ] Manual: \`roxabi issues\` from \`~/projects/roxabi-forge\` (localPath-less registered repo) — repro, verify fix
- [ ] Manual: \`roxabi issues\` from a non-git dir — verify graceful \"not registered\" message
- [ ] Manual: \`roxabi workspace add owner/repo\` from inside that repo's clone — verify \`localPath\` captured

## Note on pre-existing test failures

\`cli/__tests__/issues.test.ts\` has 5 pre-existing failing tests on \`staging\` (SC-10 / SC-11 and the \"empty table\" case). They expect batched-query output but don't pass \`all: true\`, so \`resolveCurrentProject\` errors out. Not touched by this PR — scope creep to fix. These aren't run by CI (\`cli/__tests__/**\` excluded from vitest in \`vitest.config.ts\`).


MickaelV0:
## Code Review

**Verdict:** 🟥 Request changes — 1 real blocker, 3 warnings, 2 suggestions.

_Security-auditor agent was rate-limited; this review is test-focused. Re-run security after quota resets or self-verify the subprocess/FS surface (low risk: static args, no shell, guarded JSON.parse)._

### 🛑 Blockers

**issue: `as { repo?: string }` is a compile-time cast, not runtime validation**

`plugins/dev-core/cli/lib/workspace.ts:96`

Root cause: `.roxabi` is user-authored JSON. `JSON.parse(...) as { repo?: string }` does not enforce the type. A malformed marker like `{ "repo": 42 }` passes `if (data.repo)` (truthy number), is returned as-is, and later crashes `resolveCurrentProject` at `slug.toLowerCase()` with `TypeError: slug.toLowerCase is not a function`.

Solutions:
  1. Guard with `if (typeof data.repo === 'string' && data.repo) return data.repo` (recommended)
  2. Runtime schema validation via `zod` / `valibot` (overkill for one field)

Confidence: 85%

### ⚠️ Warnings

**issue: Dynamic `import()` in every `it` body — silent trap for future stateful refactors**

`plugins/dev-core/cli/__tests__/resolve.test.ts:43,50,57,64,71,76,90,98,107,117,126,136`

Root cause: Each test calls `await import('../lib/workspace')`. Bun caches modules after first load; there is no `resetModules` between tests. For the current pure utilities this works by accident. If someone later adds module-level state (env cache, singleton), tests will silently share it.

Solutions:
  1. Hoist to top-level `import { parseGitRemoteUrl, resolveRepoFromCwd, detectLocalPath } from '../lib/workspace'` (recommended — removes the false impression of per-test isolation)
  2. Keep dynamic imports but document why + add a `resetModules` hook if/when module state appears

Confidence: 88%

**issue: `originalCwd` captured at module parse time, not per-test**

`plugins/dev-core/cli/__tests__/resolve.test.ts:176`

Root cause: `const originalCwd = process.cwd()` runs once at parse time. If any prior test/suite `chdir`s without cleanup, restoration uses a stale reference.

Solutions:
  1. Move `const originalCwd = process.cwd()` inside `beforeEach` (recommended)
  2. Wrap each `detectLocalPath` test body in `try/finally { process.chdir(originalCwd) }`

Confidence: 82%

**issue: `afterEach` not `try/finally` — `rmSync` skipped if `chdir` throws**

`plugins/dev-core/cli/__tests__/resolve.test.ts:186-191`

Root cause: `process.chdir(originalCwd)` is called before `rmSync`, unguarded. If `chdir` throws (e.g. stale cwd), the temp dir leaks.

Solutions:
  1. `try { process.chdir(originalCwd) } finally { rmSync(tmpDir, { recursive: true, force: true }) }` (recommended)
  2. Accept low risk given the unique tmpdir prefix per run

Confidence: 75%

### 💡 Suggestions (non-blocking)

**suggestion(non-blocking): Move `resolveCurrentProject` from `commands/issues.ts` to `lib/workspace.ts`**

Root cause: It was promoted to `export` only for testability. `resolveRepoFromCwd` (which it depends on) already lives in `lib/workspace.ts`. Colocating removes the command-module import side-effect surface in tests.

Defer — scope creep for this PR. Good candidate for the follow-up that consolidates `cli/lib/workspace.ts` + `skills/shared/adapters/workspace-helpers.ts`.

Confidence: 78%

**nitpick: `'returns null for unrecognized URLs'` tests two inputs in one `it` block**

`plugins/dev-core/cli/__tests__/resolve.test.ts:73-76`

Split into `'returns null for a malformed string'` and `'returns null for an empty string'` so a failure points at the specific case.

Confidence: 88%

**thought: No boundary test for walk-up at filesystem root**

`plugins/dev-core/cli/__tests__/resolve.test.ts`

The `while (dir.length > 1)` termination is untested. One-liner `expect(resolveRepoFromCwd('/')).toBeNull()` documents the contract.

Confidence: 70%

### ✨ Praise

**praise: Clean integration approach — real git subprocess + real FS, zero over-mocking**

Real `Bun.spawnSync(['git', 'init', ...])` + real `mkdtempSync`. No `vi.mock('../lib/workspace')` that would give 0% coverage. Right call for filesystem/process functions.

Confidence: 95%

---

### Summary

| Category | Count |
|---|---:|
| Blockers | 1 |
| Warnings | 3 |
| Suggestions | 3 |
| Praise | 1 |

🤖 [Claude Code](https://claude.com/claude-code) · code-review skill

MickaelV0:
## Review Fixes Applied

**Auto-applied (C≥80, verified):** 4 findings
**Applied via 1b1:** 1 finding
**Deferred (issues created):** 0
**Skipped:** 2 (praise, thought-boundary)
**Failed:** 0

### Auto-Applied (commit `71269bc`)

- **[applied]** issue(blocker): `.roxabi` type guard in `cli/lib/workspace.ts` (85%)
  `as { repo?: string }` replaced with runtime `typeof data.repo === 'string' && repo.length > 0`. Two regression tests added: non-string `{ repo: 42 }` and empty-string `{ repo: '' }` both fall through to git remote.
- **[applied]** issue: hoist dynamic test imports to top-level (88%)
  Eliminates `await import()` in every `it` body — removes the silent-trap risk if module-level state is ever added.
- **[applied]** issue: `originalCwd` / `originalHome` captured per-test (82%)
  Moved from module-parse-time `const` into `beforeEach`. `afterEach` now uses `try/finally` so `rmSync` always runs (warning #4 — partial credit applied here).
- **[applied]** nitpick: split `'returns null for unrecognized URLs'` (88%)
  One `it` per input — malformed vs empty string.

### Applied via 1b1 (commit `2d7f339`)

- **[applied]** suggestion(non-blocking): move `resolveCurrentProject` to `lib/workspace.ts` (78%)
  Generalized signature to `<P extends { repo, localPath? }>` so any project shape works. Test imports collapse to a single source.

### Skipped

- **praise:** clean integration approach (no action)
- **thought:** filesystem-root boundary test — low value, termination contract already implicit in `dir.length > 1` guard

### Test status

23/23 tests pass. Lint + typecheck clean. Pre-commit + pre-push hooks green on both commits.

🤖 [Claude Code](https://claude.com/claude-code) · fix skill

MickaelV0:
## Code Review — Round 2

**Verdict:** 🟥 Request changes — security-auditor surfaced 1 blocker + 1 todo (both low-severity but cheap to close).

### 🛑 Blockers

**issue(blocking): Unvalidated slug reflected into terminal output enables escape-sequence injection**

`plugins/dev-core/cli/lib/workspace.ts:95-102` (regex) · `plugins/dev-core/cli/commands/issues.ts:90-92` (output)

Root cause: `parseGitRemoteUrl` captures `(.+)$` — an unconstrained trailing match that admits any character, including ANSI/OSC escape sequences. The returned slug is passed to `console.log` via "Detected repo: {slug}" and "Run: roxabi workspace add {slug}" messages. An attacker who controls `.roxabi` in a directory the victim navigates to (or a poisoned `.git/config`) can inject hyperlinks or terminal control via `{"repo": "owner/\x1b]8;;evil\x07text\x1b]8;;\x07"}`. Real GitHub repo names follow `[A-Za-z0-9._-]+/[A-Za-z0-9._-]+`.

Solutions:
  1. Add allowlist in `parseGitRemoteUrl` before returning: `if (!/^[A-Za-z0-9._-]+\/[A-Za-z0-9._-]+$/.test(slug)) return null`. Apply same check in the `.roxabi` repo-field validation. (recommended)
  2. Strip non-printable / non-ASCII on output: `slug.replace(/[^\x20-\x7E]/g, '?')`.

Confidence: 72% · security-auditor

### ⚠️ Todo

**todo: `detectLocalPath` admits `..` segments that probe above `$HOME`**

`plugins/dev-core/cli/lib/workspace.ts:155-163`

Root cause: `repo.split('/')[1]` extracts the second path component without validation. For `repo = "Roxabi/.."`, candidates become `$HOME/projects/..`, `$HOME/..`, `$HOME/src/..` — `existsSync` resolves them, and any `.git` match outside the intended tree is returned as `localPath` and persisted to `workspace.json`.

Solutions:
  1. Validate `name` immediately after `split`: `if (!name || name.includes('/') || name.includes('..') || name.startsWith('.')) return undefined`. (recommended)
  2. `path.resolve` each candidate and assert it starts with `home`.

Confidence: 75% · security-auditor

### 💡 Suggestions

**suggestion(non-blocking): `.roxabi` read is unbounded — memory-pressure DoS from crafted file**

`plugins/dev-core/cli/lib/workspace.ts:105`

Root cause: `readFileSync(marker, 'utf8')` has no size limit. A 500 MB `.roxabi` planted on the walk-up path loads entirely into memory before `JSON.parse`. Requires local FS write access — rated low.

Solutions:
  1. `statSync(marker).size > 64 * 1024` → skip this marker. (recommended)
  2. Stream-parse with a bounded reader.

Confidence: 62% · security-auditor

### ✨ Praise

**praise: Subprocess call is injection-safe** (97%) — `Bun.spawnSync(['git', ...])` receives a static string array; `cwd` passed as `-C` arg, not string-concatenated.

**praise: GraphQL variables are injection-safe** (98%) — `owner`/`name` passed as JSON `variables`, never interpolated into query template.

**praise: `JSON.parse` + shape guard prevents prototype pollution** (97%) — the `typeof === 'object' && 'repo' in data && typeof repo === 'string'` chain correctly rejects malformed markers.

---

### Summary

| Category | Count |
|---|---:|
| Blockers | 1 |
| Todo | 1 |
| Suggestions | 1 |
| Praise | 3 |

All findings are low-severity (local-FS preconditions) but the slug allowlist and `..` guard are one-line fixes with real attack surface.

🤖 [Claude Code](https://claude.com/claude-code) · code-review skill (round 2)

MickaelV0:
## Review Fixes Applied — Round 2

**Auto-applied (security, first-principles verified):** 3 findings
**Skipped:** 3 (praise)
**Failed:** 0

All 3 security findings applied in a single commit (`2fd146a`) since they share the same file and are structurally related.

### Applied (commit `2fd146a`)

- **[applied]** issue(blocker): **#S1 slug escape-sequence injection** (72%)
  `parseGitRemoteUrl` captured `(.+)$` unconstrained → adversarial `.roxabi` or `.git/config` could inject ANSI/OSC escape sequences into `console.log` via the "Detected repo: {slug}" error path. Added `REPO_SLUG_RE = /^[A-Za-z0-9._-]+\/[A-Za-z0-9._-]+$/` allowlist enforced on every slug return (SSH parse, HTTPS parse, `.roxabi` validator).
- **[applied]** todo: **#S2 `detectLocalPath` path traversal** (75%)
  `repo: "Roxabi/.."` made `$HOME/projects/..` etc. probe-able via `existsSync`. Added guard: reject `name` containing `/`, `..`, or leading `.` before building candidate paths.
- **[applied]** suggestion(non-blocking): **#S3 `.roxabi` unbounded read** (62%)
  Added `statSync(marker).size <= 64KB` guard (`ROXABI_MAX_BYTES`). Oversized markers skipped silently and fall through to git remote.

### Regression tests added

| Target | Cases |
|---|---|
| `parseGitRemoteUrl` | escape sequences, path traversal, whitespace |
| `resolveRepoFromCwd` | `.roxabi` with escape sequences, `.roxabi` > 64 KB |
| `detectLocalPath` | `..`, `../etc`, `.hidden` |

29/29 tests pass. Lint + typecheck + pre-commit + pre-push hooks green.

🤖 [Claude Code](https://claude.com/claude-code) · fix skill (round 2)

MickaelV0:
## Code Review — Round 3

**Verdict:** 🟨 Approve with comments — no blockers. 4 warnings worth addressing, 2 suggestions, 2 deferred.

Tester + architect agents ran in parallel.

### ⚠️ Warnings

**issue: `detectLocalPath` traversal test cannot prove the guard fires**

`plugins/dev-core/cli/__tests__/resolve.test.ts:228-234`

Root cause: The test asserts `detectLocalPath('Roxabi/..')` returns `undefined` without planting a `.git` directory at the traversal target. On a clean tmpdir, `existsSync` returns false regardless of whether the guard exists — removing the guard entirely keeps the test green.

Solutions:
  1. Plant `.git` at the would-be traversal target so `undefined` can only come from the guard (recommended)
  2. Spy on `existsSync` to assert it is never called when the guard fires

Confidence: 88% · tester

**issue: Three `expect()` calls in one `it()` body**

`plugins/dev-core/cli/__tests__/resolve.test.ts:222-234`

Root cause: First failure hides the other two. Inconsistent with the round-2 split pattern (#9).

Solutions:
  1. Split into three named `it()` blocks: `"rejects '..' segment"`, `"rejects multi-segment traversal"`, `"rejects dot-prefixed name"` (recommended)
  2. Add assertion messages

Confidence: 82% · tester

**issue: `cli/lib/workspace.ts` has crossed the god-module threshold**

Three orthogonal concerns co-located: (1) workspace config I/O, (2) GitHub GraphQL discovery, (3) cwd/FS resolution. 193 lines, growing. Each has different failure modes and change rates.

Solutions:
  1. Split into `workspace-store.ts` / `github-discovery.ts` / `cwd-resolver.ts` (recommended — next structural PR)
  2. Add barrel re-export now, move files in a follow-up

Confidence: 88% · architect

**Deferred.** Explicit out-of-scope for this PR.

**issue: `.roxabi` marker has no schema, no documented format**

Root cause: File read + validated only against `REPO_SLUG_RE`. No documented format (JSON? line?), no version field, no structured validator. Any future format change has no migration path.

Solutions:
  1. Document format inline (JSDoc on `resolveRepoFromCwd`), extract `parseRoxabiMarker(raw): string | null`, add `version` field if it will ever grow (recommended)
  2. JSDoc-only contract + diagnostic error on malformed content

Confidence: 85% · architect

### ⚠️ Todo

**todo: File concrete follow-up for `workspace-helpers.ts` duplication**

The older `detectLocalPath` in `skills/shared/adapters/workspace-helpers.ts` lacks the round-2 traversal guard. Every PR that touches either file widens the drift. Not debt — latent security regression.

Solutions:
  1. File a concrete issue (not "some follow-up"): delete helpers file, re-export from `cli/lib/workspace.ts`, audit skill callers (recommended)
  2. Inline the traversal guard into helpers as a band-aid

Confidence: 82% · architect

### 💡 Suggestions

**suggestion(non-blocking): `REPO_SLUG_RE` is more permissive than GitHub's actual rules**

`plugins/dev-core/cli/lib/workspace.ts:5`

Root cause: `/^[A-Za-z0-9._-]+\/[A-Za-z0-9._-]+$/` admits `.owner/.name`. GitHub rejects leading dots in owner and repo names. Currently caught downstream by `detectLocalPath.startsWith('.')`, but the layer's intent diverges.

Solutions:
  1. Tighten: `/^[A-Za-z0-9][A-Za-z0-9._-]*\/[A-Za-z0-9][A-Za-z0-9._-]*$/` + test (recommended)
  2. Leave permissive, add comment documenting the intentional gap

Confidence: 75% · tester

**nitpick: `resolveCurrentProject` generic has no subtype callers in this PR**

Root cause: Generic was added in round-1 1b1. All current callers pass `WorkspaceProject[]` directly. Generic adds visual complexity without payoff.

Solutions:
  1. Replace `<P extends ...>` with concrete `WorkspaceProject` parameter
  2. Keep generic + add `@example` JSDoc to signal intent

Defer — generics cost ~nothing, real callers with subtypes will appear with the schema refactor. Confidence: 75% · architect

### ✨ Praise

- **tester:** Core security invariants covered mechanically — each guard paired with fallthrough assertion.
- **architect:** Path-traversal guard + `ROXABI_MAX_BYTES` are the right defaults for a CLI operating on untrusted workspace state.

---

### Summary

| Category | Count |
|---|---:|
| Blockers | 0 |
| Warnings | 4 |
| Todo | 1 |
| Suggestions | 2 |
| Praise | 2 |

🤖 [Claude Code](https://claude.com/claude-code) · code-review skill (round 3)

MickaelV0:
## Review Fixes Applied — Round 3

Commit `91f366d`. All 4 warnings + 1 suggestion addressed. 2 items deferred with reasoning.

### Applied

| # | Finding | What changed |
|---|---|---|
| T1 | `detectLocalPath` traversal test didn't prove guard fires | Each of 3 cases now plants the filesystem such that removing the guard flips the assertion (plant `$HOME/.git`, `$HOME/etc/.git`, `$HOME/projects/.hidden/.git` respectively) |
| T2 | Three `expect()` in one `it()` | Split into 3 named `it()` blocks — `".."`, `"../etc"`, `".hidden"` |
| T3 | `REPO_SLUG_RE` admitted leading dots | Tightened to `/^[A-Za-z0-9][A-Za-z0-9._-]*\/[A-Za-z0-9][A-Za-z0-9._-]*$/`. First char of each segment must be alphanumeric — matches GitHub's own rules. 2 new regression tests for leading-dot rejection. |
| A3 | `.roxabi` marker had no documented format | Added JSDoc to `resolveRepoFromCwd` documenting: JSON object, required `repo` string, size cap, extra fields ignored (forward-compat note), failure semantics. Light-touch — did not extract `parseRoxabiMarker` since that belongs with the god-module split. |
| A2 | `workspace-helpers.ts` duplication | Filed concrete follow-up: #101 (audit callers, delete duplicate, re-export from canonical module, grep guard in CI). |

### Deferred (with reasoning)

| # | Finding | Reasoning |
|---|---|---|
| A1 | `cli/lib/workspace.ts` god-module split | Explicit out-of-scope per PR description. Three modules (`workspace-store`, `github-discovery`, `cwd-resolver`) is the right end state but deserves its own PR. |
| A4 | `resolveCurrentProject` generic premature | Generic costs ~nothing structurally. Real subtype callers will appear with the projects-first schema refactor. Not worth reverting. |

### Test status

33/33 tests pass (+4 vs round 2: 2 leading-dot rejections, 3 provable-guard splits from 1). Lint + typecheck + pre-commit + pre-push hooks green.

🤖 [Claude Code](https://claude.com/claude-code) · fix skill (round 3)

MickaelV0:
## 📋 Cumulative Findings Summary — All 3 Review Rounds

Compiled from the /code-review + /fix cycles over this PR. 26 total findings surfaced, 12 applied, 4 deferred with reasoning, 10 praise.

### 📊 Overview

| Round | Agents | Findings | Verdict | Resolution |
|---|---|---|---|---|
| 1 | tester | 11 | 🟥 Request changes | 5 applied · 2 deferred · 4 praise/thought/question |
| 2 | security-auditor | 6 | 🟥 Request changes | 3 applied · 0 deferred · 3 praise |
| 3 | tester + architect | 9 | 🟨 Approve with comments | 4 applied + 1 deferred + 2 deferred + 2 praise |

**Verdict trend:** 🟥 → 🟥 → 🟨 · **Test growth:** 20 → 23 → 29 → 33 · **Commits:** 5 (fix + refactor + security + polish)

---

### 🛑 Blockers (all closed)

| ID | Finding | C | Agent | Round | Fix |
|---|---|---:|---|---:|---|
| R1-#5 | `as { repo?: string }` compile-time cast → runtime `TypeError` on `{repo:42}` | 85% | tester | 1 | Runtime `typeof === 'string'` guard + 2 regression tests (`71269bc`) |
| R2-#S1 | Unvalidated slug → ANSI/OSC escape-sequence injection into `console.log` | 72% | security | 2 | `REPO_SLUG_RE` allowlist enforced at every slug return (`2fd146a`) |

---

### ⚠️ Warnings — Test Hygiene (all closed)

| ID | Finding | C | Agent | Round | Fix |
|---|---|---:|---|---:|---|
| R1-#2 | Dynamic `await import()` in every `it` body — silent-trap if module-state added | 88% | tester | 1 | Hoisted to top-level imports (`71269bc`) |
| R1-#3 | `originalCwd` captured at module parse time — stale ref risk | 82% | tester | 1 | Moved into `beforeEach` (`71269bc`) |
| R1-#4 | `afterEach` not `try/finally` — `rmSync` skipped if `chdir` throws | 75% | tester | 1 | Wrapped in `try/finally` (`71269bc`) |
| R1-#9 | Two `expect()` in one `it` block | 88% | tester | 1 | Split into 2 named tests (`71269bc`) |
| R3-T1 | Traversal test didn't prove guard fires — green on any clean tmpdir | 88% | tester | 3 | Each case now plants the filesystem so removing the guard flips the assertion (`91f366d`) |
| R3-T2 | Three `expect()` in one `it()` — inconsistent with round-2 split | 82% | tester | 3 | Split into 3 named `it()` blocks (`91f366d`) |

---

### ⚠️ Warnings — Security (all closed)

| ID | Finding | C | Agent | Round | Fix |
|---|---|---:|---|---:|---|
| R2-#S2 | `detectLocalPath` admitted `"Roxabi/.."` → probed above `$HOME` | 75% | security | 2 | Guard rejects `name` containing `/`, `..`, or leading `.` (`2fd146a`) |
| R2-#S3 | `.roxabi` read was unbounded — memory-pressure DoS vector | 62% | security | 2 | `statSync().size > 64 KB` → skip (`2fd146a`) |
| R3-T3 | `REPO_SLUG_RE` admitted leading dots; GitHub forbids | 75% | tester | 3 | Tightened to `/^[A-Za-z0-9][A-Za-z0-9._-]*\/[A-Za-z0-9][A-Za-z0-9._-]*$/` + 2 regression tests (`91f366d`) |

---

### ⚠️ Warnings — Architecture (partial)

| ID | Finding | C | Agent | Round | Resolution |
|---|---|---:|---|---:|---|
| R3-A3 | `.roxabi` marker has no documented format/schema | 85% | architect | 3 | ✅ JSDoc format spec added to `resolveRepoFromCwd` (`91f366d`). Deferred full `parseRoxabiMarker` extraction to the split PR. |

---

### 💡 Accepted Suggestions

| ID | Finding | C | Agent | Round | Fix |
|---|---|---:|---|---:|---|
| R1-#6 | `resolveCurrentProject` should live in `lib/workspace.ts` (colocation) | 78% | tester | 1 | Moved + generalized to `<P extends {repo, localPath?}>` (`2d7f339`) |

---

### 📝 Todo — Filed as Follow-Up

| ID | Finding | C | Agent | Round | Action |
|---|---|---:|---|---:|---|
| R3-A2 | `workspace-helpers.ts` duplicate `detectLocalPath` lacks traversal guard = latent regression | 82% | architect | 3 | **Issue #101** filed: audit callers, consolidate, add CI grep guard |

---

### 🗓️ Deferred with Reasoning

| ID | Finding | C | Agent | Round | Why deferred |
|---|---|---:|---|---:|---|
| R1-#7 | Inconsistent fixture depth (real `git init` vs bare `mkdirSync('.git')`) | 72% | tester | 1 | Non-blocking, now moot — round-3 test rewrite uses consistent planted `.git` dirs throughout |
| R1-#10 | No boundary test for walk-up at `/` | 70% | tester | 1 | `thought` category — termination contract is implicit in `dir.length > 1`; not worth a filesystem-root test |
| R1-#11 | Bun test scheduler interleaving risk | 90% | tester | 1 | `question` category — agent self-resolved: Bun runs serially within a file, no action needed |
| R3-A1 | `cli/lib/workspace.ts` god-module — split into `workspace-store` / `github-discovery` / `cwd-resolver` | 88% | architect | 3 | Explicit out-of-scope for this PR ("fix the bug"). Separate structural PR — likely paired with #101 resolution |
| R3-A4 | `resolveCurrentProject<P extends ...>` generic has no subtype callers | 75% | architect | 3 | Generic costs ~nothing. Real subtype callers will appear with the projects-first schema refactor |

---

### ✨ Praise (10 findings)

**Round 1 (tester):**
- **Clean integration approach** — real `git init` subprocess + real `mkdtempSync`, zero over-mocking. "Right call for filesystem/process functions." (95%)

**Round 2 (security-auditor):**
- **Subprocess call is injection-safe** — static arg list, no shell interpolation, `cwd` passed as `-C` arg not concatenated (97%)
- **GraphQL variables are injection-safe** — `owner`/`name` passed as JSON `variables`, never interpolated into query template (98%)
- **`JSON.parse` + shape guard prevents prototype pollution** — correct `typeof === 'object' && 'repo' in data` narrowing chain (97%)

**Round 3 (tester):**
- **Core security invariants covered mechanically** — each guard paired with a fallthrough assertion (no confidence reported)

**Round 3 (architect):**
- **Traversal guard + `ROXABI_MAX_BYTES` are solid hardening** — right defaults for a CLI on untrusted workspace state

---

### 📦 Deliverables

| | |
|---|---|
| **Fix** | `16eeb7f` — initial resolver + `.roxabi` marker + git-remote fallback |
| **Refactor** | `2d7f339` — move `resolveCurrentProject` to lib (round 1 1b1) |
| **Hardening** | `71269bc` — typeof guard + test hygiene (round 1 auto) |
| **Security** | `2fd146a` — slug allowlist + traversal guard + size cap (round 2) |
| **Polish** | `91f366d` — provable test guards + tightened regex + docs (round 3) |
| **Follow-up** | Issue **#101** — workspace-helpers consolidation |

### 🧪 Final test state

- **33 tests pass** (up from 20 initial)
- Lint · Typecheck · Pre-commit hooks · Pre-push hooks — all ✅ across every commit
- CI green on latest push (check rollup: SUCCESS)

### 🎯 What's NOT in this PR (by design)

- **A1** — 3-way module split (`workspace-store` / `github-discovery` / `cwd-resolver`) — architect-approved refactor, deserves its own PR. Pair with #101.
- **Projects-first schema** — reshape `workspace.json` from repo-first to `{projects:[{id, repos[], fieldIds}]}`. Breaking change; discussed up-front and deferred.
- **Consolidation of `skills/shared/adapters/workspace-helpers.ts`** — tracked in #101.

---

🤖 Compiled across 3 /code-review + /fix cycles. [Claude Code](https://claude.com/claude-code)

MickaelV0:
## Code Review — Round 4

**Verdict:** 🟨 Approve with comments — no blockers. 4 warnings, 4 suggestions, 4 praise.

Scope: delta vs commit `91f366d` (the 3 round-4 commits — split + lefthook flag + test fixes). Rounds 1-3 covered the underlying logic; this round only changes packaging.

Agents: security-auditor + architect + tester + devops, parallel.

### ⚠️ Warnings

**issue: `detectLocalPath` placement leaks GitHub-discovery concern into `cwd-resolver.ts`**

`plugins/dev-core/cli/lib/cwd-resolver.ts` (last function)

Root cause: `detectLocalPath` is called exclusively by `discoverProject` in `github-discovery.ts` to enrich a freshly-fetched `WorkspaceProject` with a local clone path. Its only caller is in the GitHub-discovery concern. It ended up in `cwd-resolver.ts` because it calls `resolveRepoFromCwd` internally — but that's an implementation detail, not a semantic reason. The module boundary is now defined by internal call graph rather than by concern, the exact anti-pattern the split was meant to fix. Same finding from a second angle: the resulting `github-discovery → cwd-resolver` import pulls the network I/O layer into the FS resolution layer with no architectural justification.

Solutions:
  1. Move `detectLocalPath` into `github-discovery.ts`. `cwd-resolver.ts` then exports only the four resolution primitives forming a coherent cwd-to-repo-slug concern. (recommended)
  2. Inject `detectLocalPath` as an optional parameter to `discoverProject(repo, localPathFn = detectLocalPath)` — explicit but adds API surface.

Confidence: 84% · architect

---

**issue: pre-commit AND CI now both skip lint on `plugins/dev-core/cli/__tests__/`**

`lefthook.yml:4` + `biome.json` (`files.includes` excludes `!plugins/dev-core/cli/__tests__`)

Root cause: The new `--no-errors-on-unmatched` flag plays correctly for unblocking commits, but `biome.json files.includes` is global — it affects local, pre-commit, AND CI invocations uniformly. CI's `bun run lint` runs `biome check .` which still respects the same exclusion. Net result: 33+ test files have zero biome coverage at any stage. The hook flag was a fix for *commit ergonomics*; it inadvertently also confirmed a *quality gap that has no safety net*.

Solutions:
  1. Remove `!plugins/dev-core/cli/__tests__` from `biome.json files.includes`, run `bunx biome check --write` over the dir once to clear any accumulated noise, accept the one-time fixup. Permanently closes the gap. (recommended)
  2. Tighten the lefthook glob to `!plugins/dev-core/cli/__tests__/**` so lefthook never passes those paths to biome — the flag becomes unnecessary AND a misplaced non-test `.ts` file would still be linted.
  3. Split configs: `biome.json` (strict, used in CI) vs `biome.local.json` (excludes `__tests__`, used in lefthook).

Confidence: 91% · devops

---

**issue: `runIssuesCommand`'s default (`!opts.all`) cwd-resolution path now has zero integration test coverage**

`plugins/dev-core/cli/__tests__/issues.test.ts` (all 5 integration tests)

Root cause: The fix correctly added `all: true` to all 5 SC-10/SC-11 tests — those genuinely test the batched path. But every integration test now routes through `if (opts.all)`. The default branch (cwd → `resolveCurrentProject` → `fetchProjectItems` → format) has no integration test. `resolve.test.ts` covers the resolver helpers in isolation, but the full integration of resolver + token + HTTP + format for the single-project path is untested. Token-resolution regression, format-output regression, or error-message-format change in that branch goes uncaught.

Solutions:
  1. Add one integration test for the default path that controls cwd via a fixture project whose `localPath` matches a tmpdir — the localPath branch of `resolveCurrentProject` doesn't shell out, so it works without a real git repo. Mock fetch, assert output contains the expected label. (recommended)
  2. Accept the gap with a `// TODO: add integration test for !opts.all path` comment.

Confidence: 90% · tester

---

**issue: `readWorkspace` returns unvalidated `Workspace` — `projectId: undefined` will silently produce empty GitHub queries post-A4**

`plugins/dev-core/cli/lib/workspace-store.ts:30`

Root cause: `JSON.parse(readFileSync(p, 'utf8')) as Workspace` is a compile-time cast only. Any workspace.json on disk missing `projectId` (manually edited, written by an external tool, authored before `discoverProject` existed) deserializes to `{ projectId: undefined }` at runtime. The pre-A4 fallback `matched.projectId ?? matched.id ?? ''` papered over this. Now `fetchProjectItems(matched.projectId, token)` passes `undefined` as a `string`-typed argument, and `buildBatchedVariables` inserts `undefined` into the GraphQL variables object. The GitHub API responds 200 with a `null` node — silent failure. User sees an empty project, not an error revealing the misconfiguration. Data-integrity issue, not an exploit path.

Solutions:
  1. Add a runtime narrowing helper `parseWorkspace` (or inline in `readWorkspace`): filter projects where `typeof p.projectId !== 'string' || !p.projectId`, log the dropped entries to stderr, return only valid ones. Consistent with the `REPO_SLUG_RE` / `typeof` guards already applied to `.roxabi` parsing. (recommended)
  2. Guard at each call site in `runIssuesCommand` — lower leverage, the batched path still passes `undefined` through.

Confidence: 72% · security-auditor

### 💡 Suggestions

**suggestion(non-blocking): `IssuesCommandWorkspace` is now a structurally identical wrapper over `Workspace` — drop it**

`plugins/dev-core/cli/commands/issues.ts:11-13`

A4 deleted `IssuesCommandProject` correctly. But `IssuesCommandWorkspace { projects: WorkspaceProject[] }` was left behind, even though it is now byte-identical to `Workspace`. The wrapper had a reason when `IssuesCommandProject` differed; it has none now. Replace with `Workspace` directly in `IssuesCommandOptions.workspace?: Workspace` — `readWorkspace()` already returns `Workspace`, so no other code changes.

Confidence: 74% · architect

---

**suggestion(non-blocking): mark the barrel re-export `@deprecated` and migrate the 4 call sites**

`plugins/dev-core/cli/lib/workspace.ts:1`

The barrel comment explains *what* it does but not that it's *transitional*. Without a `@deprecated` JSDoc, maintainers will treat it as intentional API surface and accumulate new re-exports. Four call sites (commands × 2, tests × 2); migration is mechanical.

Solutions:
  1. Add `/** @deprecated Import directly from workspace-store, cwd-resolver, or github-discovery. */` above the first `export *`, plan a follow-up to migrate + delete. (recommended)
  2. Migrate the 4 call sites in this PR and delete `workspace.ts` entirely — slightly bigger diff but no transitional state.

Confidence: 78% · architect

---

**suggestion(non-blocking): the stale "RED tests" comment block in `issues.test.ts:3-12` is now misleading**

`plugins/dev-core/cli/__tests__/issues.test.ts:3-12`

The header reads "These tests will fail until the following implementations exist" — implementations exist, all 5 tests now pass. Replace with a one-line description: "Integration tests for multi-project batched GraphQL (SC-10, SC-11)."

Confidence: 95% · tester

---

**suggestion(non-blocking): rename `cwd-resolver.ts` → `repo-resolver.ts` (or document scope explicitly)**

`plugins/dev-core/cli/lib/cwd-resolver.ts`

The module exports `parseGitRemoteUrl` (pure URL parsing) and `resolveRepoFromCwd` / `resolveCurrentProject` / `detectLocalPath`. Only the latter three are cwd-related; URL parsing is a lower-level utility. The unifying concept is "derive a repo slug from various sources" — `repo-resolver` fits all four exports. Becomes more important if W1 (move `detectLocalPath`) is rejected, less important if it's accepted.

Confidence: 72% · architect

### ✨ Praise

- **security-auditor:** File split preserves every security guard from rounds 2-3 byte-for-byte (`REPO_SLUG_RE`, `ROXABI_MAX_BYTES`, `statSync` size cap, `typeof repo === 'string'`, `detectLocalPath` traversal rejection, `0o700` / `0o600` permissions). Barrel exports 11 names with zero collisions; `WorkspaceProject` is owned by exactly one module and imported as `type` by the others.
- **security-auditor:** `--no-errors-on-unmatched` is correctly scoped — only suppresses biome's exit code when zero matched files; real lint errors in non-excluded files still produce exit 1. Trufflehog (separate hook) handles secret scanning unaffected.
- **security-auditor + architect:** The `.id` fallback removal is provably safe — the field never existed in `WorkspaceProject` schema (traced back to initial CLI commit `508ae5591e`); `discoverProject` always wrote `projectId`, never `id`. Dead code, correctly deleted.
- **tester:** SC-11 ordering assertions are structurally sound — positive-existence guards on both `## frontend` / `## backend` headers (via `>= 0`) AND on issue numbers (via `toContain`) mean the silent `-1 < -1` ambiguity cannot fire under realistic mock data. Mock isolation (reassigning `fetchMock` inside the test, then `globalThis.fetch`) correctly avoids polluting `beforeEach` state.

---

### Summary

| Category | Count |
|---|---:|
| Blockers | 0 |
| Warnings | 4 |
| Suggestions | 4 |
| Praise | 4 |

🤖 [Claude Code](https://claude.com/claude-code) · code-review skill (round 4)

MickaelV0:
## Review Fixes Applied — Round 4

**Auto-applied (C≥80, 2+ agents):** 3 findings (commit `40b1b1f`)
**Applied via 1b1:** 3 findings (commit `8f7e420`)
**Deferred (issue created):** 1 finding (#103)
**Skipped:** 1 finding (reasoning below)
**Failed:** 0

### Auto-Applied — `40b1b1f`

| # | Finding | What changed |
|---|---|---|
| W2 | biome exclusion left `__tests__` unlinted at pre-commit AND CI (devops 91%, architect-confirmed 99%) | Removed `!plugins/dev-core/cli/__tests__` from `biome.json files.includes`. Ran `biome check --write` + `--unsafe` over the dir once: 6 errors + 1 warning + 1 info → 0/0/0. |
| W3 | default `!opts.all` path had zero integration coverage (tester 90%, architect-confirmed 98%) | Added one test `default path: resolves cwd via localPath and fetches a single project`. Uses `localPath: process.cwd()` so `resolveCurrentProject` matches via localPath without shelling out to git. |
| S3 | stale "RED tests" header comment (tester 95%, self-verified) | Replaced with `// Integration tests for multi-project batched GraphQL (SC-10, SC-11) plus the default single-project (\`!opts.all\`) cwd-resolution path.` |

### Applied via 1b1 — `8f7e420`

| # | Finding | What changed |
|---|---|---|
| W4 | `readWorkspace` lacked runtime validation (security 72%) | Added fail-loud `parseWorkspace(raw): Workspace`. Throws with specific field path (`workspace.json: projects[2].projectId must be a non-empty string`) so user can fix their file. Also removed the JSON-parse `try/catch` that silently returned `{ projects: [] }` on malformed JSON. 8 new tests covering happy path, missing fields, wrong types, empty strings, null input. |
| S1 | dead `IssuesCommandWorkspace` wrapper (architect 74%) | Byte-identical to `Workspace` after A4 removed `IssuesCommandProject`. Deleted the interface, use `Workspace` directly in `IssuesCommandOptions.workspace`. Removed the `as IssuesCommandWorkspace` cast. |
| S2 | `workspace.ts` barrel re-export (architect 78%) | Chose the alternative (solution 2) over `@deprecated`: deleted `workspace.ts` entirely and migrated the 4 call sites to direct imports of `workspace-store` / `cwd-resolver` / `github-discovery`. No transitional state, zero ambiguity about which module owns what. |

### Deferred — [#103](https://github.com/Roxabi/roxabi-plugins/issues/103)

**W1** — `detectLocalPath` placement (architect 84%, round-4 verifier REJECTED at 82%)

The round-4 verifier (security-auditor) rejected the architect's proposed fix because:
1. `detectLocalPath` has a second caller — `resolve.test.ts` imports it directly, grouped with the other cwd-resolution primitives in the test suite.
2. Moving the function to `github-discovery.ts` doesn't clean up the dependency graph — `github-discovery` would still need to import `resolveRepoFromCwd` from `cwd-resolver`.

The root problem is the **heuristic `$HOME` scan** (`~/projects/<name>`, `~/<name>`, `~/src/<name>`), which is a hardcoded layout convention that only works for users who follow that filesystem pattern. It is also the reason the function ended up in `cwd-resolver.ts` in the first place (the cwd-match branch and the HOME-scan branch have different concerns glued together).

Filed [#103](https://github.com/Roxabi/roxabi-plugins/issues/103) to remove the HOME scan, add a `--local <path>` flag to `roxabi workspace add`, and reconsider the module placement once the function is simplified. That's its own PR with UX implications (new flag + argument parsing + tests).

### Skipped

**S4** — rename `cwd-resolver.ts` → `repo-resolver.ts` (architect 72%)

The rename's motivation was that the module contains `parseGitRemoteUrl` (pure URL parsing, no cwd semantics) plus three cwd-related functions. Once #103 lands and `detectLocalPath`'s HOME scan is gone, every function in the module really does resolve something *about the cwd*, and `parseGitRemoteUrl` is just a helper used exclusively by `resolveRepoFromCwd`. Renaming now adds churn for marginal gain; the structural fix is #103, not a file rename.

### Final state

| Gate | Status |
|---|---|
| CLI tests | **62/62 pass** (was 53, +9 new: 1 integration + 8 parseWorkspace) |
| Full suite | 399/405 (same baseline — 4 pre-existing `scaffoldFumadocs` fails, unrelated; 2 `.todo`) |
| Typecheck | ✅ |
| Biome | ✅ (now covers `__tests__/` too) |
| Pre-commit + pre-push hooks | ✅ |

🤖 [Claude Code](https://claude.com/claude-code) · fix skill (round 4)

PR #84, opened 2026-04-09
feat(forge): add base/ layer + shells/ for design system

MickaelV0:
## Summary
- Extract forge's prose-embedded CSS/JS from `tokens.md` and `split-file.md` into 8 actual code files (`base/` × 6 + `shells/` × 2) so skills read and inline verbatim — eliminates per-invocation CSS re-derivation and visual drift
- Introduces semantic color token contract (`--success/warning/error/info` + dim variants) and anti-pattern enforcement block in `components.css`

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #78: forge: foundation — base/ layer + shells/ (S1–S3) | Open |
| Analysis | [forge-design-system-evolution-analysis.mdx](artifacts/analyses/forge-design-system-evolution-analysis.mdx) | Present (epic-level) |
| Spec | [78-forge-foundation-base-shells-spec.mdx](artifacts/specs/78-forge-foundation-base-shells-spec.mdx) | Approved |
| Implementation | 1 commit on `feat/78-forge-foundation-base-shells` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (296 passed, 0 new) | Passed |

## Files

| File | Purpose |
|------|---------|
| `base/reset.css` | Box-sizing, antialiasing, Lyra dark/light token blocks |
| `base/layout.css` | Nav, tabs, panels, cards grid, responsive breakpoint |
| `base/typography.css` | IBM Plex fonts, type scale, heading/paragraph/code styles |
| `base/components.css` | Cards, tables, semantic color tokens, anti-pattern block |
| `base/tab-loader.js` | `loadPanel()`/`activate()` with `{NAME}` placeholder |
| `base/mermaid-init.js` | `__postLoad` + `__initPanZoom` (inline verbatim) |
| `shells/single.html` | Self-contained diagram shell (`file://` safe) |
| `shells/split.html` | Multi-tab shell with tab-loader injection slot |

## Test Plan
- [ ] `cat base/reset.css base/layout.css base/typography.css base/components.css` produces valid concatenated CSS
- [ ] All hex color values appear only in `:root`/`[data-theme]` token blocks — no hardcoded colors in component rules
- [ ] `shells/single.html` contains all 13 placeholders from spec
- [ ] `shells/split.html` contains all 15 placeholders, theme toggle appears before `{TAB_LOADER_JS}`
- [ ] `gallery-templates/` directory is unchanged (zero modifications)
[...]

Issue #79, opened 2026-04-09
forge: lyra + roxabi aesthetics (S4)


MickaelV0:
## What

Create `references/aesthetics/lyra.css` and `references/aesthetics/roxabi.css`.

Extract current brand tokens from tokens.md into actual CSS files. Each file defines:
- `:root` token overrides (bg, surface, border, text tiers, accent)
- Semantic color overrides (--success, --warning, --error, --info + dim)
- Font pairing (lyra: IBM Plex Sans + Mono; roxabi: same with gold accent)
- Dark mode as default

### lyra.css
- Dark bg (#0a0a0f), orange accent (#e85d04)
- 3-tier text: --text (18:1), --text-muted (8:1), --text-dim (4.3:1)

### roxabi.css
- Dark bg (#111210), gold accent (#f0b429)
- Same text hierarchy

## Acceptance

- [ ] Both files define the full token contract from base/components.css
- [ ] forge-chart output for lyra project uses lyra.css
- [ ] forge-chart output for roxabi-plugins uses roxabi.css
- [ ] No token values hardcoded — all via CSS variables

PR #129, opened 2026-04-22
fix(dev-core): #116 backport quality-gate shell fixes (macOS bash compat + awk exemption)

### LINKED ISSUE #116: dev-core: backport quality-gate shell fixes from voiceCLI #79 (macOS bash compat + awk exemption match)
## Summary

voiceCLI PR https://github.com/Roxabi/voiceCLI/pull/84 (closes voiceCLI#79) applied two fixes to its project-side copy of the dev-core quality-gate shell scripts. Both need to be backported to the canonical sources in \`plugins/dev-core/tools/\`, otherwise the next \`/release-setup --force\` on any consumer repo will overwrite the fixes with the unfixed canonical versions.

## Fix 1 — macOS bash 3.2 compatibility

**File:** \`plugins/dev-core/tools/check_folder_size.sh\`

**Current:** uses \`mapfile -d '' files < <(find ... -print0)\` which requires bash ≥4.4. macOS ships bash 3.2.57 and won't update (GPLv2), so contributors on macOS without Homebrew bash cannot run the hook (\`bash: mapfile: command not found\` under \`set -euo pipefail\`).

**Fix applied in voiceCLI PR #84:**

\`\`\`bash
# Portable NUL-delimited count — works on macOS bash 3.2 (no mapfile/readarray).
COUNT=0
while IFS= read -r -d '' _; do
    COUNT=$((COUNT + 1))
done < <(find "$d" -maxdepth 1 -name "*.py" -type f -print0)
\`\`\`

Matches the NUL-delimited idiom already used in the outer \`find | while read\` loop.

## Fix 2 — left-anchored exemption match (awk field-split)

**Files:** \`plugins/dev-core/tools/check_file_length.sh\` + \`plugins/dev-core/tools/check_folder_size.sh\`

**Current:** \`grep -qF -- "\$1 " "\$EXEMPT_FILE"\` — fixed-string substring search with trailing-space right-anchor only. Not exploitable with current exemption entries, but a future exemption URL containing a path-like substring with a trailing space could theoretically false-positive. security-auditor flagged it at 62% confidence.

**Fix applied in voiceCLI PR #84:**

\`\`\`bash
is_exempt() {
    [ ! -f "\$EXEMPT_FILE" ] && return 1
    # Exact match on the first whitespace-delimited field — no regex, no escaping,
    # left-anchored (awk field split) so a path substring elsewhere on the line
    # cannot cause a false positive. Exemption format: '<path> <issue-url>'.
    awk -v p="\$1" '\$1 == p { found = 1 } END { exit !found }' "\$EXEMPT_FILE"
}
\`\`\`

Awk field match is exact on \`\$1\`. No regex escaping needed for paths with metacharacters. Left-anchored by definition.

## Acceptance

- [ ] Both fixes landed in \`plugins/dev-core/tools/\` canonical sources
- [ ] Release cut so consumer repos can \`/release-setup --force\` to pull the fixes without reintroducing the bugs
- [ ] Other consumer repos (if any) with drifted copies are identified and re-stamped

## Reference

- voiceCLI PR: https://github.com/Roxabi/voiceCLI/pull/84
- voiceCLI issue: https://github.com/Roxabi/voiceCLI/issues/79
- Cookbook: \`plugins/dev-core/skills/release-setup/cookbooks/quality-gates.md\`

MickaelV0:
## Summary

Backports two quality-gate shell fixes from [voiceCLI PR #84](https://github.com/Roxabi/voiceCLI/pull/84) (closes voiceCLI#79) into the canonical `plugins/dev-core/tools/` sources. Without these, the next `/release-setup --force` on any consumer repo would silently re-introduce the bugs.

## Fix 1 — macOS bash 3.2 compat (`check_folder_size.sh`)

Replace `mapfile -d ''` (bash ≥4.4) with a portable `while IFS= read -r -d '' _; do COUNT=$((COUNT+1)); done` counter. macOS ships bash 3.2.57 (GPLv2, frozen), so contributors without Homebrew bash hit `bash: mapfile: command not found` under `set -euo pipefail`.

## Fix 2 — left-anchored exemption match (both scripts)

Replace `grep -qF -- "$1 " "$EXEMPT_FILE"` with `awk -v p="$1" '$1 == p { found = 1 } END { exit !found }'`. Exact match on the first whitespace-delimited field — no regex, no escaping, left-anchored by definition. Closes the theoretical false-positive where an exemption line containing a path substring with trailing space could match unrelated paths (security-auditor flagged at 62% confidence on voiceCLI #79).

## Test plan

- [x] `bash -n` syntax check on both scripts
- [x] Smoke test on isolated repo: 3 .py files, max=2 — exempt path passes, non-exempt fails with correct count
- [ ] CI green
- [ ] After merge: cut release so consumer repos can `/release-setup --force` to pull the fixes

Closes #116

### DIFF
--- a/plugins/dev-core/tools/check_file_length.sh
+++ b/plugins/dev-core/tools/check_file_length.sh
@@ -25,9 +25,10 @@ fi
 
 is_exempt() {
     [ ! -f "$EXEMPT_FILE" ] && return 1
-    # grep -F (fixed-string): path may contain regex metacharacters (., *, [, +).
-    # Exemption format: '<path> <issue-url>' — trailing space anchors the path.
-    grep -qF -- "$1 " "$EXEMPT_FILE"
+    # Exact match on the first whitespace-delimited field — no regex, no escaping,
+    # left-anchored (awk field split) so a path substring elsewhere on the line
+    # cannot cause a false positive. Exemption format: '<path> <issue-url>'.
+    awk -v p="$1" '$1 == p { found = 1 } END { exit !found }' "$EXEMPT_FILE"
 }
 
 while IFS= read -r -d '' f; do
--- a/plugins/dev-core/tools/check_folder_size.sh
+++ b/plugins/dev-core/tools/check_folder_size.sh
@@ -24,16 +24,19 @@ fi
 
 is_exempt() {
     [ ! -f "$EXEMPT_FILE" ] && return 1
-    # grep -F (fixed-string): path may contain regex metacharacters (., *, [, +).
-    # Exemption format: '<path> <issue-url>' — trailing space anchors the path.
-    grep -qF -- "$1 " "$EXEMPT_FILE"
+    # Exact match on the first whitespace-delimited field — no regex, no escaping,
+    # left-anchored (awk field split) so a path substring elsewhere on the line
+    # cannot cause a false positive. Exemption format: '<path> <issue-url>'.
+    awk -v p="$1" '$1 == p { found = 1 } END { exit !found }' "$EXEMPT_FILE"
 }
 
 while IFS= read -r -d '' d; do
     is_exempt "$d" && continue
-    # -print0 + mapfile keeps the array consistent with the outer loop's NUL-delimited pattern.
-    mapfile -d '' files < <(find "$d" -maxdepth 1 -name "*.py" -type f -print0)
-    COUNT=${#files[@]}
+    # Portable NUL-delimited count — works on macOS bash 3.2 (no mapfile/readarray).
+    COUNT=0
+    while IFS= read -r -d '' _; do
+        COUNT=$((COUNT + 1))
+    done < <(find "$d" -maxdepth 1 -name "*.py" -type f -print0)
     if [ "$COUNT" -gt "$MAX" ]; then
         echo "$d - $COUNT files (max $MAX)"
         FAIL=1


MickaelV0:
## Code Review

**Verdict:** Approve with comments — 1 non-blocking finding (72% C). Mechanical backport from already-reviewed voiceCLI PR #84; both fixes verified by smoke test + green CI + green `/validate`.

### Suggestions

**suggestion(non-blocking): `awk -v` silently consumes backslashes in path strings**
- `plugins/dev-core/tools/check_file_length.sh:31`
- `plugins/dev-core/tools/check_folder_size.sh:30`
- _-- security-auditor_
- **Root cause:** POSIX awk's `-v var=value` processes escape sequences in the assigned value before storing it. A backslash followed by a recognised escape letter (`\b`, `\t`, `\n`) is converted to the corresponding control character; an unrecognised pair drops the backslash. A Linux path containing a literal `\n` (e.g. `src/foo\nbar.py` — legal on ext4 but extremely rare for Python sources) would arrive in `p` with the two chars replaced by a newline, so `$1 == p` never matches and the exemption is silently ignored. **Impact: false negative** (operator's exemption ignored — the script flags a file the operator intended to exempt). No false-positive, no security bypass.
- **Solutions:**
  1. Use `ENVIRON` to bypass `-v` escape processing: `P="$1" awk '$1 == ENVIRON["P"] { found=1 } END { exit !found }' "$EXEMPT_FILE"` (recommended).
  2. Accept the limitation with a comment noting paths containing backslashes are unsupported by the exemptions mechanism (pragmatic — rare in Python source trees).
- **Confidence:** 72%
- **Note:** This finding applies equally to the upstream voiceCLI #84. Fixing here without fixing upstream would re-create drift this PR was opened to eliminate. Recommend deferring to a follow-up issue that lands the `ENVIRON` form in *both* repos simultaneously.

### Praise

- **praise:** Smoke-tested against an isolated repo before commit (3 .py files, max=2: exempt path passes, non-exempt fails with correct count). Test plan in PR body matches the local verification.
- **praise:** Commit message and PR body both link upstream voiceCLI PR #84 + voiceCLI#79, preserving the audit trail across repos.

### Vectors checked, no findings

- Shell injection: `$1` and `$EXEMPT_FILE` are passed as awk arguments, never interpolated into an evaluated string. No regression vs. prior `grep -qF`.
- `set -e` × awk non-zero exit: safe — `is_exempt` is called as `is_exempt "$d" && continue`, condition position is exempt from `set -e` abort.
- 0-byte exemption file: awk handles correctly (no records, exits 1 = not exempt).
- `COUNT` subshell scoping: process substitution `< <(...)` preserves parent shell scope; counter increment visible after the loop. Correct.
- Pre-existing `read -r` truncation-at-newline for paths with embedded newlines: not introduced by this PR.

PR #147, opened 2026-05-05
fix(dev-core): replace awk -v with ENVIRON in exemption check

### LINKED ISSUE #130: dev-core: harden quality-gate awk exemption against backslash-bearing paths (cross-repo: roxabi-plugins + voiceCLI)
## Summary

Code-review on PR #129 (backport of voiceCLI #84) flagged a follow-up at 72% confidence: the `awk -v p="$1" '$1 == p { found=1 } END { exit !found }' "$EXEMPT_FILE"` exemption-match pattern in `plugins/dev-core/tools/check_file_length.sh` and `check_folder_size.sh` silently consumes backslash escape sequences in the assigned value.

A path like `src/foo\nbar.py` (legal on ext4, extremely rare for Python sources) arrives in awk's variable `p` with `\n` replaced by a literal newline → `\$1 == p` never matches → the operator's exemption is silently ignored.

**Impact:** false negative (operator-intended exemption ignored). No false positive, no security bypass.

## Fix

Replace `-v` with `ENVIRON` to bypass awk's escape processing:

\`\`\`bash
P=\"\$1\" awk '\$1 == ENVIRON[\"P\"] { found = 1 } END { exit !found }' \"\$EXEMPT_FILE\"
\`\`\`

## Cross-repo coordination

The exact same idiom exists in voiceCLI's project-side copy (introduced by voiceCLI PR #84). Fixing only one side reintroduces the drift that #116 was opened to eliminate.

**Required:**

- [ ] Apply `ENVIRON` fix to `plugins/dev-core/tools/check_file_length.sh` + `check_folder_size.sh` here.
- [ ] Apply identical fix to voiceCLI's project-side copies (open companion voiceCLI issue + PR).
- [ ] Land both before the next dev-core release tag so consumer repos pull a consistent canonical version on \`/release-setup --force\`.

## Reference

- PR #129 review comment: https://github.com/Roxabi/roxabi-plugins/pull/129#issuecomment-4295798496
- voiceCLI PR (origin): https://github.com/Roxabi/voiceCLI/pull/84
- Parent backport issue: #116

MickaelV0:
## Summary

- Replace \`awk -v p=\"\$1\"\` with \`P=\"\$1\" awk '... ENVIRON[\"P\"] ...'\` in \`check_file_length.sh\` and \`check_folder_size.sh\` — fixes silent exemption bypass for paths with backslash sequences
- Extend \`biome.json\` file exclusions to cover all \`.venv\` dirs (was only \`plugins/web-intel/.venv\`) and \`.claude/worktrees\` — eliminates nested-config lint failure

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #130: harden quality-gate awk exemption against backslash-bearing paths | Open |
| Implementation | 2 commits on \`fix/130-awk-environ-exemption\` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests 988/988 ✅ | Passed |

## Test Plan

- [ ] Verify \`is_exempt\` returns true for a path containing \`\\\n\` in the exemption file
- [ ] Verify normal paths still match correctly in both \`check_file_length.sh\` and \`check_folder_size.sh\`
- [ ] Run \`bun run lint\` — should show 0 errors (worktree + .venv exclusions take effect)

## Cross-repo

The same awk fix is needed in voiceCLI's project-side copies (companion issue required — noted in #130).

Closes #130

---

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -20,7 +20,7 @@
     }
   },
   "files": {
-    "includes": ["**/*.ts", "**/*.js", "!node_modules", "!**/.venv", "!.cache", "!.claude/worktrees"]
+    "includes": ["**/*.ts", "**/*.js", "!**/node_modules", "!**/.venv", "!**/.cache", "!.claude/worktrees"]
   },
   "overrides": []
 }
--- a/biome.json
+++ b/biome.json
@@ -20,7 +20,7 @@
     }
   },
   "files": {
-    "includes": ["**/*.ts", "**/*.js", "!node_modules", "!plugins/web-intel/.venv", "!.cache", "!.claude/worktrees"]
+    "includes": ["**/*.ts", "**/*.js", "!node_modules", "!**/.venv", "!.cache", "!.claude/worktrees"]
   },
   "overrides": []
 }
--- a/plugins/dev-core/tools/check_file_length.sh
+++ b/plugins/dev-core/tools/check_file_length.sh
@@ -28,7 +28,8 @@ is_exempt() {
     # Exact match on the first whitespace-delimited field — no regex, no escaping,
     # left-anchored (awk field split) so a path substring elsewhere on the line
     # cannot cause a false positive. Exemption format: '<path> <issue-url>'.
-    awk -v p="$1" '$1 == p { found = 1 } END { exit !found }' "$EXEMPT_FILE"
+    # ENVIRON avoids awk's -v escape processing (backslash sequences in path → corrupted match).
+    P="$1" awk '$1 == ENVIRON["P"] { found = 1 } END { exit !found }' "$EXEMPT_FILE"
 }
 
 while IFS= read -r -d '' f; do
--- a/plugins/dev-core/tools/check_file_length.sh
+++ b/plugins/dev-core/tools/check_file_length.sh
@@ -28,9 +28,16 @@ is_exempt() {
     # Exact match on the first whitespace-delimited field — no regex, no escaping,
     # left-anchored (awk field split) so a path substring elsewhere on the line
     # cannot cause a false positive. Exemption format: '<path> <issue-url>'.
+    # Paths must not contain spaces — field splitting would break the match.
     # ENVIRON avoids awk's -v escape processing (backslash sequences in path → corrupted match).
     P="$1" awk '$1 == ENVIRON["P"] { found = 1 } END { exit !found }' "$EXEMPT_FILE"
 }
+
+# Guard: exemption paths must not contain spaces (NF>2 means path has embedded whitespace).
+if [ -f "$EXEMPT_FILE" ] && awk 'NF > 2 { exit 0 } END { exit 1 }' "$EXEMPT_FILE"; then
+    echo "ERROR: $EXEMPT_FILE: exemption path contains spaces — paths with spaces are not supported" >&2
+    exit 1
+fi
 
 while IFS= read -r -d '' f; do
     is_exempt "$f" && continue
--- a/plugins/dev-core/tools/check_file_length.sh
+++ b/plugins/dev-core/tools/check_file_length.sh
@@ -33,8 +33,9 @@ is_exempt() {
     P="$1" awk '$1 == ENVIRON["P"] { found = 1 } END { exit !found }' "$EXEMPT_FILE"
 }
 
-# Guard: exemption paths must not contain spaces (NF>2 means path has embedded whitespace).
-if [ -f "$EXEMPT_FILE" ] && awk 'NF > 2 { exit 0 } END { exit 1 }' "$EXEMPT_FILE"; then
+# Guard: exemption paths must not contain spaces (NF>2 on a non-comment line means embedded whitespace).
+# Skips comment lines (#) to avoid false-positives on scaffold-generated exemption file headers.
+if [ -f "$EXEMPT_FILE" ] && awk '/^[[:space:]]*#/ { next } NF > 2 { found=1 } END { exit !found }' "$EXEMPT_FILE"; then
     echo "ERROR: $EXEMPT_FILE: exemption path contains spaces — paths with spaces are not supported" >&2
     exit 1
 fi
--- a/plugins/dev-core/tools/check_folder_size.sh
+++ b/plugins/dev-core/tools/check_folder_size.sh
@@ -32,8 +32,9 @@ is_exempt() {
     P="$1" awk '$1 == ENVIRON["P"] { found = 1 } END { exit !found }' "$EXEMPT_FILE"
 }
 
-# Guard: exemption paths must not contain spaces (NF>2 means path has embedded whitespace).
-if [ -f "$EXEMPT_FILE" ] && awk 'NF > 2 { exit 0 } END { exit 1 }' "$EXEMPT_FILE"; then
+# Guard: exemption paths must not contain spaces (NF>2 on a non-comment line means embedded whitespace).
+# Skips comment lines (#) to avoid false-positives on scaffold-generated exemption file headers.
+if [ -f "$EXEMPT_FILE" ] && awk '/^[[:space:]]*#/ { next } NF > 2 { found=1 } END { exit !found }' "$EXEMPT_FILE"; then
     echo "ERROR: $EXEMPT_FILE: exemption path contains spaces — paths with spaces are not supported" >&2
     exit 1
 fi
--- a/plugins/dev-core/tools/check_folder_size.sh
+++ b/plugins/dev-core/tools/check_folder_size.sh
@@ -27,9 +27,16 @@ is_exempt() {
     # Exact match on the first whitespace-delimited field — no regex, no escaping,
     # left-anchored (awk field split) so a path substring elsewhere on the line
     # cannot cause a false positive. Exemption format: '<path> <issue-url>'.
+    # Paths must not contain spaces — field splitting would break the match.
     # ENVIRON avoids awk's -v escape processing (backslash sequences in path → corrupted match).
     P="$1" awk '$1 == ENVIRON["P"] { found = 1 } END { exit !found }' "$EXEMPT_FILE"
 }
+
+# Guard: exemption paths must not contain spaces (NF>2 means path has embedded whitespace).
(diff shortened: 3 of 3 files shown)


MickaelV0:
## Code Review

### Blockers

**issue**: Exemption format silently fails for paths containing spaces
  `plugins/dev-core/tools/check_file_length.sh:32` · `plugins/dev-core/tools/check_folder_size.sh:31`
  -- devops
  Root cause: `awk` field-splits on whitespace, so `$1` of `src/my module/foo.py` is `src/my` — match fails silently even after the ENVIRON fix. The ENVIRON change is correct but orthogonal to field-split semantics. Linux Python paths rarely have spaces, but the format contract is unenforced.
  Solutions:
    1. Add a guard before the while loop: `if grep -q ' ' "$EXEMPT_FILE" 2>/dev/null; then echo "ERROR: spaces in exempt path" >&2; exit 1; fi` + document in exemption file header (recommended)
    2. Switch to tab delimiter + `FS="\t"` in awk to support space-in-path
  Confidence: 80%

---

**suggestion**: `!node_modules` does not cover `**/node_modules` subdirectories
  `biome.json:23`
  -- devops
  Root cause: Biome 2.x `files.includes` uses glob semantics — `!node_modules` anchors to project root only; nested `node_modules` would still be scanned.
  Solutions:
    1. Change `"!node_modules"` → `"!**/node_modules"` (recommended)
    2. Add `"!**/node_modules/**"` alongside existing pattern
  Confidence: 72%

---

### Nitpicks

**nitpick**: Biome negation patterns mix glob depths — `!.cache` and `!.claude/worktrees` lack `**` prefix per Biome 2.x docs
  `biome.json:23`
  -- devops
  Root cause: Biome docs state negations should be prefixed with `**` when `**/*.ts`/`**/*.js` positive patterns are present. Works empirically today but may break on future Biome patch.
  Solutions:
    1. Use `["**", "**/*.ts", "**/*.js", "!**/node_modules", "!**/.venv", "!**/.cache", "!.claude/worktrees"]`
    2. Add CI smoke-test asserting `.claude/worktrees` stays excluded
  Confidence: 65%

---

### Praise

**praise**: `P="$1" awk '$1 == ENVIRON["P"]'` correctly sidesteps `awk -v` backslash-escape processing
  `plugins/dev-core/tools/check_file_length.sh:32` · `plugins/dev-core/tools/check_folder_size.sh:31`
  -- devops
  POSIX-compliant, survives paths with `\n`/`\t`, identical fix applied consistently across both scripts.
  Confidence: 95%

---

### Summary

| Finding | C | Blocks |
|---------|---|--------|
| Exemption format: spaces in paths not guarded | 80% | ✓ |
| `!node_modules` not covering subdirs | 72% | ✗ |
| Biome negation glob depth inconsistency | 65% | ✗ |
| ENVIRON fix is correct | 95% | praise |

**Verdict: Request changes** — 1 blocker (spaces-in-paths guard, trivial to add). No security concerns.

MickaelV0:
## Review Fixes Applied

**Auto-applied (C≥80, 2+ agents):** 0 finding(s)
**Applied via 1b1:** 3 finding(s)
**Deferred (issues created):** 0
**Skipped:** 0
**Failed:** 0

### Applied (1b1)

- [applied] issue(80%): Exemption format silently fails for paths with spaces — added `NF > 2` guard before the main loop in `check_file_length.sh` and `check_folder_size.sh`; documented format constraint in comment
- [applied] suggestion(72%): `!node_modules` not covering subdirs — changed to `!**/node_modules` in `biome.json`
- [applied] nitpick(65%): Biome negation glob depth inconsistency — changed `!.cache` to `!**/.cache` in `biome.json`

### Bonus fix

- [applied] fix(test): excluded `.claude/worktrees/**` from vitest root discovery — worktree test files were being picked up by the pre-push hook, causing unrelated failures

MickaelV0:
## Code Review (iteration 2)

### Blockers

**issue**: awk `exit 0` in rule body is shadowed by `END { exit 1 }` — spaces guard is dead code, never fires
  `plugins/dev-core/tools/check_file_length.sh:37` · `plugins/dev-core/tools/check_folder_size.sh:36`
  -- security-auditor + devops
  Root cause: POSIX awk always executes the END block even after `exit expr` in a rule body; if END contains its own `exit`, that exit code wins. `awk 'NF > 2 { exit 0 } END { exit 1 }'` exits 1 on every input — the guard condition is never true.
  Solutions:
    1. Replace with flag pattern: `awk 'NF > 2 { found=1 } END { exit !found }'` — END runs once, exits correctly (recommended)
    2. Use grep: `grep -qP '^[^ ]+ [^ ]+ ' "$EXEMPT_FILE"`
  Confidence: 98%

**issue**: NF > 2 guard does not skip comment lines — false-positive on standard exemption file headers
  `plugins/dev-core/tools/check_file_length.sh:37` · `plugins/dev-core/tools/check_folder_size.sh:36`
  -- security-auditor
  Root cause: Scaffold-generated exemption files include header comments like `# Each entry MUST reference a tracking issue` (NF=13). Once the dead-code bug is fixed, any exemption file with these headers would trigger the guard and abort the check — breaking all repos using the standard scaffold.
  Solutions:
    1. Skip comment lines in guard: `awk '/^[[:space:]]*#/ { next } NF > 2 { found=1 } END { exit !found }'` (recommended)
    2. Grep with comment exclusion: `grep -v '^[[:space:]]*#' "$EXEMPT_FILE" | grep -qP '^[^ ]+ [^ ]+ '`
  Confidence: 85%

---

**suggestion**: `!.claude/worktrees` in biome.json lacks `/**` — relies on biome-internal dir-prefix semantics
  `biome.json:23`
  -- devops
  Root cause: Works in biome 2.4.5 via biome-specific implicit directory exclusion, but not standard glob behaviour — other tools require `/**` suffix to exclude directory contents.
  Solutions:
    1. Change to `!.claude/worktrees/**` (recommended)
    2. Keep with comment documenting the biome-specific dependency
  Confidence: 80%

---

### Nitpicks

**nitpick**: vitest `.claude/worktrees/**` exclude should use `**/` prefix for cwd-independence
  `vitest.config.ts:10`
  -- devops + tester
  Root cause: Relative pattern works when vitest is run from repo root (always true today), but diverges from `**/node_modules/**` / `**/cli/__tests__/**` convention in the same file. `**/.claude/worktrees/**` is safe from any cwd.
  Solutions:
    1. Change to `**/.claude/worktrees/**` (recommended)
    2. Add comment documenting root-anchored intent
  Confidence: 72%

**nitpick**: No test coverage for `check_file_length.sh` / `check_folder_size.sh`
  `plugins/dev-core/tools/check_file_length.sh` · `plugins/dev-core/tools/check_folder_size.sh`
  -- tester
  Root cause: Logic changes (ENVIRON fix, spaces guard) are untested. A regression in `is_exempt` wouldn't be caught by CI.
  Solutions:
    1. Add Vitest integration test using `Bun.spawnSync` (matching `licenseChecker.test.ts` pattern) — cover exit codes for compliant run, file-over-limit, FIND_ROOT absent, spaces-in-exemption guard
    2. Add bats test suite under `tools/__tests__/`
  Confidence: 88%

---

### Summary

| Finding | C | Blocks |
|---------|---|--------|
| awk guard dead code (exit shadowed) | 98% | ✓ |
| NF guard false-positive on comment lines | 85% | ✓ |
| biome worktree pattern lacks `/**` | 80% | ✗ |
| vitest pattern cwd-dependence | 72% | ✗ |
| No shell script test coverage | 88% | ✗ |

**Verdict: Request changes** — 2 blockers in the spaces guard implementation.

MickaelV0:
## Review Fixes Applied (iteration 2)

**Auto-applied (C≥80, 2+ agents):** 1 finding
**Applied via 1b1:** 1 finding
**Skipped:** 1 (biome `/**` — reverted; biome ≥2.2 warns against it, bare form is correct)
**Failed:** 0

### Auto-Applied

- [applied] issue(98%): awk guard dead code — replaced `NF > 2 { exit 0 } END { exit 1 }` with flag pattern `NF > 2 { found=1 } END { exit !found }` in both scripts

### Applied (1b1)

- [applied] issue(85%): NF guard false-positive on comment lines — added `/^[[:space:]]*#/ { next }` skip at start of awk program

### Skipped

- suggestion(80%): `!.claude/worktrees/**` — biome ≥2.2 warns about trailing `/**` on folder ignores; bare `!.claude/worktrees` is the canonical form per `useBiomeIgnoreFolder` lint rule

### Verified

All 4 guard edge cases pass:
- Normal 2-field exempt file → no error ✓
- Path with space → error detected ✓  
- Empty file → no error ✓
- Comment header (NF>2) + valid data lines → no error ✓

MickaelV0:
## Code Review (iteration 3 — final)

**suggestion**: `NF > 2` guard has two edge-case gaps — trailing comments (false-positive) and missing-URL entries (false-negative)
  `plugins/dev-core/tools/check_file_length.sh:38` · `plugins/dev-core/tools/check_folder_size.sh:37`
  -- devops + security-auditor
  Root cause: `NF > 2` is a proxy for "path has a space". It fires when NF=3+ which catches space-in-path+URL, but also fires on valid entries with trailing inline comments (`src/foo.py https://... # note` → NF=4, false positive). It misses space-in-path with no URL (`src/foo bar.py` → NF=2, false negative). A tighter check would test `$2 !~ /^https?:\/\//` to verify the second field is a URL.
  Solutions:
    1. Replace `NF > 2 { found=1 }` with `NF >= 2 && $2 !~ /^https?:\/\// { found=1 }` — fires when field 2 is not a URL (path fragment or comment fragment), handles both gaps
    2. Drop the guard entirely — `is_exempt` already fails silently for spaced paths since `$1` never matches a space-containing `ENVIRON["P"]`; the guard adds a UX improvement but not correctness
  Confidence: 82%

---

### Notes

- `!.claude/worktrees` in biome.json: bare form is correct for biome ≥ 2.2 — the `useBiomeIgnoreFolder` lint rule explicitly warns against `/**` suffix. Not a finding.
- `P="$1" awk` env prefix: POSIX-correct under bash shebang. No defect in current usage (never sourced). Confidence 62% — below threshold.

---

### Summary

| Finding | C | Blocks |
|---------|---|--------|
| `NF > 2` trailing-comment false-positive + missing-URL false-negative | 82% | ✗ |

**Verdict: Approve with comments** — no blockers. One suggestion on a non-critical guard edge case.

PR #179, opened 2026-05-18
feat(dev-core): runtime-aware worktree_setup scaffolding

### LINKED ISSUE #178: dev-core:init doesn't scaffold worktree_setup hook for Python/uv projects → broken Pyright in every new worktree
## Problem

On any Python+uv project using git worktrees (i.e. every dev-core project per global rule §5), every new worktree starts with **broken Pyright imports** for third-party deps (`pydantic`, `nats`, project packages, etc.) until the user manually runs `uv sync` (~30s + disk bloat) or symlinks `.venv` from main.

Root cause: `git worktree add` doesn't copy/link `.venv`. Pyright resolves from `<cwd>/.venv` and finds nothing.

## Why the hook exists but doesn't help

`/implement` already interpolates a post-create hook at Step 2e (`skills/implement/SKILL.md:126`):

```bash
# Optional: {commands.worktree_setup} <N>
```

But the field is invisible to anyone running `/init`:

| Layer | Should expose `worktree_setup` | Today |
|---|---|---|
| `stack.yml.example` | Yes, commented stub | absent |
| `stack-setup` Phase 4 (template emission) | Yes, conditional on `runtime=python` | emits same 7 keys regardless of runtime |
| `env-setup` Phase 1 | Yes, ask when Python picked | no question |
| `/implement` Step 2e | Consumes it | only place it appears, marked `# Optional` |

`stack-setup` Phase 2 *correctly detects* `runtime=python pm=uv` (line 39–40) but Phase 4 ignores that signal.

## Proposed fix

When `stack-setup` detects `runtime=python` AND `package_manager=uv`:

1. **Scaffold `tools/worktree-setup.sh`** with the safe symlink pattern (validated on lyra — see ref below). The script must:
   - Skip if `.venv` is a real directory (preserves diverged-deps branches)
   - Replace stale symlinks
   - Skip when run inside the main repo (defensive)

2. **Register the hook** in the generated `stack.yml`:
   ```yaml
   commands:
     # ...
     worktree_setup: tools/worktree-setup.sh
   ```

3. **Echo to user** what was scaffolded so they understand the indirection.

## Files to change in dev-core

- `plugins/dev-core/stack.yml.example` — add `worktree_setup` (and symmetrical `worktree_teardown`) under `commands:` with a Python-specific comment
- `plugins/dev-core/skills/stack-setup/SKILL.md` Phase 4 — add the runtime-conditional emission + script scaffolding
- `plugins/dev-core/skills/env-setup/SKILL.md` Phase 1 — surface the hook when Python runtime is selected (optional, but improves discoverability)

## Reference implementation

Validated on lyra 2026-05-18:
- `~/projects/lyra/tools/worktree-setup.sh` — 3-branch safety logic (skip real dir / replace symlink / create fresh)
- `~/projects/lyra/.claude/stack.yml` — `commands.worktree_setup: tools/worktree-setup.sh`

## Out of scope

- Other runtimes (Node/`node_modules`, Rust/`target`) — open separate issues if/when those projects hit the same pain
- A general "post-worktree-create hook event" in Claude Code harness — `commands.worktree_setup` already covers it via dev-core

MickaelV0:
## Summary
- Adds a checklist-driven worktree-setup scaffolder to dev-core: `/init` detects runtime + package manager, then writes `tools/worktree-setup.sh` + `tools/worktree-teardown.sh` tailored to the project (uv symlink for Python, `bun install` + lefthook hooksPath fix for Bun, etc.) and registers both hooks in `.claude/stack.yml`.
- Determinism is enforced by composition: the LLM picks + orders concerns from `references/worktree-setup-checklist.md` (static bash snippets) — it can not rewrite snippet bodies, which keeps the generated scripts reproducible + shellcheck-clean.
- Includes `/env-setup` retrofit (Phase 1c) for projects with σ but no hook.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | [#178](https://github.com/Roxabi/roxabi-plugins/issues/178) | open |
| Frame | [artifacts/frames/178-worktree-setup-scaffold-frame.mdx](artifacts/frames/178-worktree-setup-scaffold-frame.mdx) | approved |
| Spec | [artifacts/specs/178-worktree-setup-scaffold-spec.mdx](artifacts/specs/178-worktree-setup-scaffold-spec.mdx) | approved |
| Plan | [artifacts/plans/178-worktree-setup-scaffold-plan.mdx](artifacts/plans/178-worktree-setup-scaffold-plan.mdx) | 11 tasks, 3 waves |
| Implementation | 2 commits on `feat/178-worktree-setup-scaffold` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (19 new) | Passed |

## What changed
- `plugins/dev-core/references/worktree-setup-checklist.md` (new) — 6 seed concerns: env-files, uv-venv-symlink, bun-install-warmup, npm-install-warmup, lefthook-hookspath-fix, neon-db-branch. Plus a safety contract + extension guide.
- `plugins/dev-core/tools/worktreeScaffold.ts` (new) — zero-dep Bun script: `parseChecklist`, `selectConcerns`, `composeScript`, `shouldOfferRetrofit`, CLI (`parse | compose | list-selected`).
- `plugins/dev-core/tools/__tests__/worktreeScaffold.test.ts` (new) — 8 unit tests (parse, select for python/uv and bun/neon contexts, compose structure, shellcheck-clean optional).
- `plugins/dev-core/tools/__tests__/scaffold-integration.test.ts` (new) — 11 integration tests across python.uv, bun.monorepo+neon, env-setup retrofit detection. Tests scrub `GIT_*` env vars + pin `GIT_CONFIG_GLOBAL` to `/dev/null` so tmp-repo git ops don't get hijacked by parent git context (e.g. pre-push hook).
- `plugins/dev-core/skills/stack-setup/SKILL.md` — new Phase 4b: runtime gate, preview, DP(A) Write/Show diff/Abort, idempotent re-run.
- `plugins/dev-core/skills/env-setup/SKILL.md` — new Phase 1c: retrofit path for existing projects (σ exists but no hook).
- `plugins/dev-core/stack.yml.example` — documents `commands.worktree_setup` + `worktree_teardown` keys.

## Test Plan
- [ ] `bun test plugins/dev-core/tools/__tests__/` — 19 tests green (8 unit + 11 integration). shellcheck-dependent tests soft-skip when binary absent.
- [ ] Run `/stack-setup` on a fresh Python+uv repo — verify Phase 4b preview shows `env-files, uv-venv-symlink, lefthook-hookspath-fix` and DP(A) gate fires before write.
- [ ] After Phase 4b writes scripts, run `git worktree add ../wt-x -b feat/x` then run the generated `tools/worktree-setup.sh` inside the worktree — `.venv` should be a symlink to the main repo's `.venv`.
- [ ] Re-run `/stack-setup` — should silent-skip when scripts present; `--force` should fire the DP(A) Regenerate/Keep/Abort prompt.
- [ ] On a project with existing σ (no hook), run `/env-setup` — Phase 1c should detect the gap and prompt.
- [ ] Validate the script's main-checkout guard: run the generated script inside the main repo (not a worktree) — should exit 0 without modifying the real `.venv`.

## Notable design call
The reviews (architect + product-lead) flagged LLM non-determinism as the deepest risk. Resolved by moving composition out of the SKILL.md (which the LLM interprets) into a pure-function TS helper. The LLM is now bounded to: presenting previews, handling DP(A) responses, and invoking the CLI. Snippets in the checklist are static bash — `shellcheck -S warning` clean by construction.

Closes #178

---


MickaelV0:
## Code Review

6-agent fresh-context review: security-auditor, architect, product-lead, tester, backend-dev, devops. 21 files, 1869 diff lines. Single chunk.

---

### Verdict: **Request changes** (5 blockers, 17 advisory, 6 praise)

The feature is well-designed and the TS pure-function approach holds up. The blockers are predominantly in the SKILL.md wizard path — the TS helper is solid but the wizard that wraps it has missing wiring + a structural cross-skill delegation that won't actually execute as written.

---

### 🚫 Blockers (5)

**B1 — `issue(blocking):` Phase 4b builds `CTX_JSON` from shell variables that Phase 2 never sets**
`plugins/dev-core/skills/stack-setup/SKILL.md:263–276`
-- product-lead + devops
Root cause: Phase 2 auto-discover uses `echo` to stdout (lines 33–98) — it does not assign shell variables. Phase 4b Step 3 then interpolates `${RUNTIME}`, `${PM}`, `${MONOREPO_BOOL:-false}`, `${HOOKS_TOOL:-lefthook}`, `${DATABASE:-none}`, `${BE_PATH}` into the heredoc. All are unset → `CTX_JSON` emits `{"runtime":"", "package_manager":"", ...}` → `selectConcerns` returns `[]` → `${COUNT}==0` early-exits every run. SC10's `runtime=bun database=neon` path is **only** reachable via the TS unit tests, not via the actual `/stack-setup` wizard. **This is the highest-severity bug in the PR — the feature does not work end-to-end as wired.**
Class: `parallel-path-drift`
Solutions:
  1. Convert Phase 2 echo-outputs to variable assignments (`RUNTIME=$(grep ... pyproject.toml | ...)`) so Phase 4b can consume them. (recommended)
  2. Have Phase 4b re-parse Phase 2's stdout block, but this duplicates detection logic.
  3. Add Phase 3 confirmation questions ("hooks tool?", "database?") so missing signals are user-supplied.
Confidence: 92%

**B2 — `issue(blocking):` `CTX_JSON` heredoc is injection-vulnerable**
`plugins/dev-core/skills/stack-setup/SKILL.md:265–276`
-- security-auditor
Root cause: Double-quoted heredoc (`<<EOF`, not `<<'EOF'`) with raw shell-variable interpolation. Values containing `"`, `\`, newline corrupt the JSON structure. Values containing `` ` ``, `$(...)`, or `${...}` are shell-expanded **before** JSON formation. `RUNTIME`/`PM` come from filename detection — they're not user-arbitrary today but the pattern is brittle and any future Phase 2 enhancement that derives values from `pyproject.toml` *content* (e.g., project name) would directly expose this surface.
Class: `shell-injection`
Solutions:
  1. Build JSON via `jq -n --arg`: `CTX_JSON=$(jq -n --arg runtime "$RUNTIME" --arg pm "$PM" --argjson monorepo "${MONOREPO_BOOL:-false}" '{runtime:$runtime, package_manager:$pm, monorepo:$monorepo, ...}')` (recommended)
  2. Use `<<'EOF'` heredoc with literal placeholders, then `sed` per-field with `jq -Rn --arg v "$VAR" '$v'` for each substitution.
Confidence: 85%

**B3 — `issue(blocking):` Cross-skill delegation is a structural smell that won't compose at runtime**
`plugins/dev-core/skills/env-setup/SKILL.md` Phase 1c step 5
-- architect + devops + product-lead (3-agent consensus)
Root cause: Step 5 says "run Phase 4b steps 3–8 of `/stack-setup`." Skills don't compose by reference — the LLM executing `/env-setup` reads only its own SKILL.md; Phase 4b text is not in context. The current Phase 1c block *does* inline the relevant `bun "${Φ}/tools/worktreeScaffold.ts" ...` commands, so the delegation phrase is just confusing prose that will silently rot if Phase 4b step numbering shifts.
Class: `parallel-path-drift`
Solutions:
  1. Drop the cross-reference text. Inline the concrete commands (CTX build → `list-selected` → DP(A) → compose+write) verbatim in Phase 1c. (recommended)
  2. Extract shared scaffolder workflow into `${CLAUDE_PLUGIN_ROOT}/references/worktree-setup-workflow.md` and have both skills `@`-include or reference it.
Confidence: 90%

**B4 — `issue(blocking):` Shellcheck tests are false-greens (vitest counts as passing, not skipped)**
`plugins/dev-core/tools/__tests__/worktreeScaffold.test.ts:141–143`
`plugins/dev-core/tools/__tests__/scaffold-integration.test.ts:111, 155, 265` (analogous `gitAvailable`/`bunAvailable` early-returns)
-- tester + product-lead
Root cause: `if (!shellcheckAvailable) return` (or analogous) inside `it(...)` body makes vitest report the test as **passed** with zero assertions. SC6 ("compose output passes shellcheck -S warning") is the contract — on a machine without shellcheck, the test silently no-ops while the report says ✓. Additionally `expect(result).toBeDefined()` is tautological since `execSync` returns a string (truthy) on exit-0 and throws otherwise — the actual gate is the `execSync` throw.
Class: `test-tautology`
Solutions:
  1. Replace with module-level `const shellcheckAvailable = (() => { try { execSync('command -v shellcheck'); return true } catch { return false } })()` then `it.skipIf(!shellcheckAvailable)('...', ...)`. Same for `gitAvailable`/`bunAvailable`. Vitest then reports honestly as "skipped". (recommended)
  2. Replace tautological assertion with `expect(result).toBe('')` (asserts shellcheck emitted no warnings).
Confidence: 85%
Raw callsites: [{file: plugins/dev-core/tools/__tests__/worktreeScaffold.test.ts, line: 141}, {file: plugins/dev-core/tools/__tests__/worktreeScaffold.test.ts, line: 168}, {file: plugins/dev-core/tools/__tests__/scaffold-integration.test.ts, line: 111}, {file: plugins/dev-core/tools/__tests__/scaffold-integration.test.ts, line: 138}, {file: plugins/dev-core/tools/__tests__/scaffold-integration.test.ts, line: 155}, {file: plugins/dev-core/tools/__tests__/scaffold-integration.test.ts, line: 177}, {file: plugins/dev-core/tools/__tests__/scaffold-integration.test.ts, line: 224}, {file: plugins/dev-core/tools/__tests__/scaffold-integration.test.ts, line: 265}]

**B5 — `issue(blocking):` `CLEAN_ENV` not applied to one scaffolder invocation in integration tests**
`plugins/dev-core/tools/__tests__/scaffold-integration.test.ts:277`
-- tester
Root cause: `execSync(\`bun "${SCAFFOLD_TS}" compose ...\`)` lacks `env: CLEAN_ENV`. Every other git-touching `execSync` carries it. A parent process with `GIT_DIR` set (the exact scenario that polluted the branch during this PR's development) would re-introduce subprocess hijacking through this one call.
Class: `parallel-path-drift`
Solutions:
  1. Add `env: CLEAN_ENV` to the options object on this `execSync`. (recommended)
Confidence: 95%

---

### ⚠️ Advisory (17)

**A1 — `suggestion(non-blocking):` `parseChecklist` doesn't validate schema version**
`plugins/dev-core/tools/worktreeScaffold.ts:44–50`
-- architect | Root cause: `version` is parsed but never compared against a `SUPPORTED_VERSION` constant. A `version: 2` checklist would parse identically with no migration path. Add `if (version !== 1) throw …` or a `// TODO: version check` so the deferred decision is visible.
Confidence: 75%

**A2 — `issue:` `parseInlineArray` corrupts apostrophes in array values**
`plugins/dev-core/tools/worktreeScaffold.ts:66`
-- backend-dev + architect | `trimmed.replace(/'/g, '"')` is a blunt instrument. A value like `["it's-fine"]` becomes invalid JSON. Safe today (no concern uses apostrophes), latent for tomorrow. Switch to a proper inline-array tokeniser or import `yaml`.
Class: `candidate/yaml-parser-fragility`
Confidence: 78%

**A3 — `issue:` `parseBlockScalar` doesn't handle CRLF line endings or tab indentation**
`plugins/dev-core/tools/worktreeScaffold.ts:76`
-- backend-dev | CRLF: `line === ''` fails for `\r` lines, terminating the block early. Tabs: 4-space indent check misses tab-indented snippets, returning empty body silently. Normalize CRLF→LF at file ingest; throw on tab indentation.
Class: `candidate/yaml-parser-fragility`
Confidence: 80%

**A4 — `issue:` Missing input validation on `$N` in `neon-db-branch` snippet**
`plugins/dev-core/references/worktree-setup-checklist.md:87–93`
-- security-auditor | `N="${1:-}"` then `bun run db:branch:create --force "$N"`. Shell quoting closes the direct injection, but the unvalidated value propagates to the user-authored DB branch script. Add `[[ "$N" =~ ^[a-zA-Z0-9_/-]{1,100}$ ]] || exit 1` guard.
Class: `missing-input-validation`
Confidence: 72%

**A5 — `issue:` SC4 sub-conditions (a) skip-real-dir and (b) replace-stale-symlink have no tests**
`plugins/dev-core/tools/__tests__/scaffold-integration.test.ts`
-- tester | Only sub-condition (c) "exit-in-main-checkout" is tested. Deleting either guard from the snippet would not be caught.
Confidence: 80%

**A6 — `issue:` SC5 teardown is composed and shellcheck-verified but never executed against a worktree**
`plugins/dev-core/tools/__tests__/scaffold-integration.test.ts`
-- tester | Teardown is the safety-critical direction. A broken `rm` glob would ship undetected. Add at minimum a test that runs the teardown script against a worktree-with-symlink and asserts post-state.
Confidence: 78%

**A7 — `suggestion(non-blocking):` env-setup Phase 1c step 4 silently mutates σ without DP(A)**
`plugins/dev-core/skills/env-setup/SKILL.md` Phase 1c step 4
-- product-lead | When script exists but σ key absent, the wizard appends the key without an affordance. Inconsistent with the preview-gate philosophy elsewhere. Either emit a one-line notice + auto-write, or wrap in a minimal DP(A) **Add keys** | **Skip**.
Confidence: 75%

**A8 — `suggestion(non-blocking):` "Show diff" branch is dead UX on the first-write path**
`plugins/dev-core/skills/stack-setup/SKILL.md` Phase 4b Step 5
-- product-lead | `diff -u <(compose) tools/worktree-setup.sh` only meaningful when the file already exists. On first-write the right hand of the diff is missing. Gate Show-diff to `--force` Regenerate path only, or rename to "Preview script" and dump the composed body.
Confidence: 88%

**A9 — `issue:` `execSync` shell strings interpolate `${ctxJson}` with single-quote wrapping; breaks on JSON containing apostrophes**
`plugins/dev-core/tools/__tests__/scaffold-integration.test.ts:136, 177, 278`
-- security-auditor | Latent — fixtures are safe today, but the pattern will be copied. Switch to `spawnSync('bun', [SCAFFOLD_TS, ...args])` to eliminate shell-quoting entirely.
Class: `shell-injection`
Confidence: 68%

**A10 — `issue:` `composeScript` noop-teardown detection is too strict**
`plugins/dev-core/tools/worktreeScaffold.ts:199`
-- backend-dev + architect | `body.trim() === ':'` misses `: # noop` or `:\n`. Either tighten the predicate (`/^:\s*(#.*)?$/`) or document the exact sentinel in the safety contract.
Confidence: 75%

**A11 — `issue:` `shouldOfferRetrofit` regex doesn't handle quoted YAML values or inline comments**
`plugins/dev-core/tools/worktreeScaffold.ts:219`
-- backend-dev | `/^runtime:\s*(\S+)/m` captures `"python"` (with quotes) for `runtime: "python"`, then `.includes('python')` returns true but only by accident. Strip quotes + trailing comment: `/^runtime:\s*["']?([\w-]+)["']?/m`.
Confidence: 70%

**A12 — `suggestion(non-blocking):` `selectConcerns` silently swallows unknown keys (hides typos)**
`plugins/dev-core/tools/worktreeScaffold.ts:163`
-- backend-dev | A checklist with `runtim=python` (typo) would silently exclude the concern. Emit `console.warn` to stderr for unknown keys without breaking selection.
Confidence: 70%

**A13 — `issue:` `parseContextJson` performs unchecked cast to `ProjectContext`**
`plugins/dev-core/tools/worktreeScaffold.ts:246`
-- backend-dev | `JSON.parse(raw) as ProjectContext` — any missing field surfaces as a runtime error downstream (`ctx.env_files.length` throws). At minimum a comment noting the unchecked cast; ideally a runtime field-presence check.
Confidence: 65%

**A14 — `suggestion(non-blocking):` Phase 4 σ template doesn't include `worktree_setup`/`worktree_teardown` as commented placeholders**
`plugins/dev-core/skills/stack-setup/SKILL.md` Phase 4 template
-- devops | When Phase 4b is skipped (unsupported runtime), generated σ doesn't show the keys at all, but `stack.yml.example` does — inconsistent. Add commented placeholders to the Phase 4 template too.
Confidence: 65%

**A15 — `issue:` Phase 1c step 4 only probes the exact filename `tools/worktree-setup.sh` — alternate names lead to duplicate scripts**
`plugins/dev-core/skills/env-setup/SKILL.md` Phase 1c step 4
-- devops | Projects with `tools/setup-worktree.sh` or `scripts/worktree-setup.sh` would land a second file. Either widen the probe (`grep 'worktree' tools/*.sh`) or warn before writing.
Confidence: 70%

**A16 — `nitpick:` `bun.lockb` test fixtures are 0 bytes — idempotency assertion trivially holds**
`plugins/dev-core/tools/__tests__/__fixtures__/projects/{bun-monorepo,retrofit-bun}/bun.lockb`
-- tester | A 0-byte lockfile means `bun install` regenerates from scratch — both runs are first runs. Real idempotency would need a minimal valid lockfile. Document the limitation or generate `bun.lock` (text format) via a fixture-prep step.
Confidence: 70%

**A17 — `nitpick:` `retrofit-bun` σ fixture is stripped down — doesn't represent realistic pre-existing dev-core project**
`plugins/dev-core/tools/__tests__/__fixtures__/projects/retrofit-bun/.claude/stack.yml`
-- tester | Missing `backend`, `frontend`, `build`, `testing`, `hooks`, `deploy`, `docs`, `artifacts` sections. Currently fine because `shouldOfferRetrofit` only checks `worktree_setup:`/`runtime:`, but masks any future check that examines other keys.
Confidence: 60%

---

### ⭐ Praise (6)

- `praise:` Pure-function design (`parseChecklist`, `selectConcerns`, `composeScript`) + `import.meta.main` CLI gating — clean separation, individually unit-testable. -- architect
- `praise:` All snippets consistently double-quote variable expansions; `set -euo pipefail` at script top; safety contract is honoured everywhere. -- security-auditor
- `praise:` `CLEAN_ENV` env-scrub + `GIT_CONFIG_GLOBAL=/dev/null` is the right pattern for git-isolated tests — applied thoroughly (one gap noted in B5). -- tester + security-auditor
- `praise:` `lefthook-hookspath-fix` guard logic (`GIT_DIR != GIT_COMMON_DIR`) is exactly correct — main-checkout case is a clean no-op. -- devops
- `praise:` Walkthrough D ("extending the checklist") has a copyable worked example + the `neon-db-branch` entry doubles as a second instance. SC2 is fully satisfied. -- product-lead
- `praise:` Style matches repo conventions (single quotes, no semicolons, clean section dividers, no `any` leaks). -- backend-dev

---

### Summary

| Severity | Count |
|---|---:|
| 🚫 Blocking | 5 |
| ⚠️ Advisory | 17 |
| ⭐ Praise | 6 |

**Verdict: Request changes.** B1–B3 are the load-bearing blockers (the feature can't ship end-to-end without B1 in particular). B4–B5 are test quality / regression-prevention concerns that should land in the same iteration to avoid CI lying about coverage. The 17 advisory findings are all genuine but can be batched.

---
Generated by `/code-review` (dev-core) — 6 agents, single chunk, 1869 diff lines

MickaelV0:
## Review Fixes Applied

**Auto-applied (pre-approved batch):** 10 finding(s)
**Deferred (next iteration):** 7 finding(s)
**Skipped:** 0
**Failed:** 0

### Auto-Applied — Blockers (5)

- [applied] **B1** `issue(blocking):` Phase 4b context-build — Phase 2 echoes are display-only; added explicit re-detect block in Phase 4b that assigns `RUNTIME`/`PM`/`HOOKS_TOOL`/`DATABASE`/`ENV_FILES`/`MONOREPO_BOOL`/`BE_PATH` before building CTX_JSON. *(plugins/dev-core/skills/stack-setup/SKILL.md)*
- [applied] **B2** `issue(blocking):` `CTX_JSON` heredoc injection — replaced with `jq -n --arg/--argjson` builder; added `command -v jq` guard. *(same file)*
- [applied] **B3** `issue(blocking):` Cross-skill delegation — dropped "delegate to /stack-setup Phase 4b steps 3-8" prose in env-setup Phase 1c; inlined the full concrete procedure. *(plugins/dev-core/skills/env-setup/SKILL.md)*
- [applied] **B4** `issue(blocking):` Shellcheck/git/bun availability gates were silent passes — promoted to module-level `const`s + `it.skipIf(!available)(...)` so vitest reports skipped, not passed; replaced tautological `toBeDefined()` with `toBe('')`. *(both test files)*
- [applied] **B5** `issue(blocking):` Missing `env: CLEAN_ENV` — refactored all 3 compose invocations into a shared `runCompose()` helper using `spawnSync` + args array (also closes A9). *(scaffold-integration.test.ts)*

### Auto-Applied — Advisory (5)

- [applied] **A2** `issue:` `parseInlineArray` — dropped naive quote rewrite; strict `JSON.parse` with descriptive error. *(worktreeScaffold.ts)*
- [applied] **A3** `issue:` Parser — normalize CRLF→LF; throw on tab indentation with the offending line in the error. *(worktreeScaffold.ts)*
- [applied] **A4** `issue:` `neon-db-branch` — added `^[a-zA-Z0-9_/-]{1,100}$` guard on `$N` in both setup + teardown snippets; updated `validation:` to document the constraint. *(worktree-setup-checklist.md)*
- [applied] **A5** `issue:` SC4 sub-conditions — added 2 integration tests: `skips when .venv exists as real dir` + `replaces stale symlink pointing nowhere`. *(scaffold-integration.test.ts)*
- [applied] **A6** `issue:` SC5 teardown — added `teardown removes .venv symlink but leaves real .venv untouched` integration test. *(same)*
- [applied] **A8** `suggestion:` Show-diff dead UX — gated to `--force` Regenerate path; fresh-install now shows "Preview composed body" via `head -80`. *(stack-setup SKILL.md)*
- [applied] **A9** `issue:` Shell-string interpolation — folded into B5's `spawnSync` refactor. *(scaffold-integration.test.ts)*
- [applied] **A10** `issue:` `composeScript` noop detection — relaxed to `/^:\s*(#.*)?\s*$/`. *(worktreeScaffold.ts)*
- [applied] **A11** `issue:` `shouldOfferRetrofit` regex — strip quotes + comments; return `true` (permissive) on unparseable. *(worktreeScaffold.ts)*
- [applied] **A15** `issue:` Phase 1c step 4 — added alternate-name probe (`grep -rl 'worktree[-_]setup' tools/*.sh scripts/*.sh`) before exact-filename check. *(env-setup SKILL.md)*

### Deferred to next iteration (7)

These advisory findings (C ≤ 75% from single agents) were not bundled in this iteration to keep the diff focused on blockers + 2-agent advisory. None are blocking the spec — track for a follow-up sweep:

- **A1** `suggestion:` Schema version not validated against `SUPPORTED_VERSION`
- **A7** `suggestion:` Phase 1c step 4 silently mutates σ without DP(A) confirmation
- **A12** `suggestion:` `selectConcerns` silently swallows unknown keys (could be a `console.warn` for typo discoverability)
- **A13** `issue:` `parseContextJson` unchecked cast — minor; runtime errors are clear
- **A14** `suggestion:` Phase 4 σ template lacks commented `worktree_setup` placeholders
- **A16** `nitpick:` `bun.lockb` test fixtures are 0 bytes (idempotency trivially satisfied)
- **A17** `nitpick:` `retrofit-bun` σ fixture is stripped down

### Verification

```
Lint        ✅ pass
Typecheck   ✅ pass
Tests       ✅ 558 passed | 4 skipped (was 559 silent-pass; net +4 new safety tests)
```

Commit: `69328ad`

---
Generated by `/fix` (dev-core)

PR #256, opened 2026-06-02
fix(quality-gates): qg.conf assignments respect caller env overrides

### LINKED ISSUE #255: fix(quality-gates): qg.conf unconditional assignments clobber env overrides in check_file_length.sh
## Problem
The canonical `plugins/dev-core/tools/check_file_length.sh` sources `tools/qg.conf` with `. tools/qg.conf`. qg.conf (generated by release-setup cookbook N4a) uses **unconditional** assignments:
```
QG_FILE_ROOT=src/
QG_FILE_EXEMPTIONS=tools/file_exemptions.txt
```
Sourcing therefore **overwrites** any env var the caller exported *before* invoking the script. Intended precedence is env > qg.conf > hardcoded default, but the implementation yields qg.conf > env.

## Impact
Repos wire package-level hooks by env override, e.g. lyra `.pre-commit-config.yaml`:
```
entry: bash -c "QG_FILE_ROOT=packages/roxabi-nats/src/ QG_FILE_EXEMPTIONS=packages/roxabi-nats/tools/file_exemptions.txt tools/check_file_length.sh"
```
qg.conf clobbers `QG_FILE_ROOT` back to `src/` → these hooks silently re-scan the main `src/` instead of the package src → package file-length checks are no-ops. Affects every repo using dev-core with package-level file-length hooks (lyra: roxabi-nats, roxabi-contracts).

## Evidence (verified)
```
$ QG_FILE_ROOT=packages/roxabi-nats/src/ bash -c '. tools/qg.conf; echo $QG_FILE_ROOT'
src/        # expected: packages/roxabi-nats/src/
```

## Fix
Emit conditional-assignment form in cookbook N4a so env wins:
```
: "${QG_FILE_MAX:=300}"
: "${QG_FILE_ROOT:=src/}"
: "${QG_FILE_EXEMPTIONS:=tools/file_exemptions.txt}"
...
```
Apply to all QG_* keys in the generated qg.conf heredoc. Both canonical scripts keep their `${VAR:-default}` reads unchanged and env overrides take effect. Note: project-side drift gates mirroring N4a (e.g. lyra `scripts/check-qg-conf-drift.sh`) must update to expect the conditional form.

## Context
Found while implementing roxabi-plugins#254 (polyglot SLOC metric). Documented in Roxabi/roxabi-factory#1672. Kept out of Roxabi/roxabi-factory#254/#1672 scope because changing qg.conf form must be coordinated with the drift gate to stay stable across /release-setup runs.

MickaelV0:
## Summary
- N4a cookbook generated `tools/qg.conf` with **unconditional** assignments (`QG_FILE_ROOT=src/`), clobbering any `QG_*` env var the caller exported before invoking a canonical script — inverting precedence to qg.conf > env.
- Impact: package-level hooks that scope a script by env override (e.g. `QG_FILE_ROOT=packages/foo/src/ tools/check_file_length.sh`) silently re-scanned the main `src/` tree → package file-length checks were no-ops.
- Fix: emit each of the 9 keys as `: "${VAR:=default}"` (conditional assignment) so an exported env var wins. Canonical scripts already read `${VAR:-default}`, unchanged → net precedence becomes **env > qg.conf > script default**. The `: "..."` wrapper also subsumes the multi-word `QG_FILE_EXTS` quoting concern.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #255: qg.conf unconditional assignments clobber env overrides | Open |
| Implementation | 1 commit on `feat/255-qg-conf-clobber-env-overrides` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (849 passed) | Passed |

## Test Plan
- [x] `QG_FILE_ROOT=packages/x/src/ bash -c '. tools/qg.conf; echo $QG_FILE_ROOT'` → prints the package path (was `src/`)
- [x] No env set → defaults still apply (`QG_FILE_ROOT=src/`, `QG_FILE_MAX=300`)
- [x] Multi-word `QG_FILE_EXTS` sources without word-splitting

## Notes
- Doc-only change to the generator cookbook; canonical `check_*.sh` scripts unchanged.
- Project-side drift gates mirroring N4a (e.g. lyra `scripts/check-qg-conf-drift.sh`) must update to expect the conditional form (out-of-repo, tracked separately).

Fixes #255

---


MickaelV0:
## Code Review

Fresh-agent review (security-auditor + devops). Diff: 1 file, doc-only (N4a generator cookbook). Single chunk, no secrets, no spec (S-tier).

**Verdict: ✅ Approve with comments** — no blocking findings.

**suggestion (devops, C 97%):** Prose `quality-gates.md:223` says `:=` "assigns only when the variable is unset" — `${VAR:=default}` also assigns when the var is **null/empty**, not unset-only. Recommend: "assigns when the variable is unset or empty". Implication: a caller exporting an empty string (`QG_FILE_ROOT=`) is overwritten by qg.conf, so the unqualified "env wins" is precise only for non-empty values. Class: `generator-drift` (doc↔semantics).

**nitpick (devops, C 85%):** Same root cause — empty-value export edge case. The PR use-case (`QG_FILE_ROOT=packages/foo/src/`) is non-empty and unaffected; empty-export-as-sentinel is not a documented pattern. Low impact.

**suggestion (security-auditor, C 62%):** The `:=` default-evaluation form makes the stack.yml → shell value path explicit; a `$(...)`/backtick in a stack.yml-derived value would be evaluated when qg.conf is sourced. **Pre-existing and equivalent** — the old `VAR=value` form embedded the same string identically; not a regression. Trust boundary sound: `.claude/stack.yml` is repo-controlled and read via `yaml.safe_load`. Optional hardening: whitelist-validate generator values (`^[A-Za-z0-9 ./_-]+$`) before the heredoc. Class: `shell-injection` (advisory).

### Praise
- env > qg.conf > default precedence verified across all three cases (env / qg.conf-only / neither).
- `: "..."` wrapper correctly subsumes multi-word `QG_FILE_EXTS` quoting; `\$` escaping keeps var refs literal while interpolating stack.yml defaults.

Fix planned: apply the C-97% doc-precision tweak (#1), then merge.

Issue #260, opened 2026-06-08
Drop status:* label sync from issue-triage set (issues-only: status = open/closed)


MickaelV0:
`issue-triage set --status` mirrors the status onto a `status:*` label via `syncStatusLabel` (`skills/issue-triage/lib/set.ts`, the label-sync block alongside size/priority/lane). In the **issues-only** model (no ProjectV2 board) status is just **open/closed**, and the dep-graph derives ready/blocked/done from edges — so `status:*` labels are redundant. Repos without these labels also get a noisy `'status:Done' not found` warning on every close/triage.

### Acceptance
- Remove the `syncStatusLabel` call from `set` (and the helper if it becomes unused)
- **Keep** the board-side `--status` write (gated on `isProjectConfigured()`) — repos still on a board are unaffected
- **Keep** `syncSizeLabel` / `syncPriorityLabel` / `syncLaneLabel` — those labels exist and are used
- Update `__tests__/set.test.ts` fixtures/expectations

### Scope
~10–20 lines, 1 core file (`set.ts`) + tests. Status becomes label-free; size/priority/lane label sync untouched.

### Related
Part of the ProjectV2/board-decoupling line:
- #253 — repo-centric CLI issues
- #252 — drop dashboard ProjectV2 dependency
- consumer: Roxabi/roxabi-live#118 (board config removed; this warning surfaced there)

Issue #252, opened 2026-06-01
Drop ProjectV2 dashboard dependency


MickaelV0:
The dashboard (`skills/issues/dashboard.ts`) is a ProjectV2 board viewer — it queries `ProjectV2.items` and reads `fieldValues` (Status, Size, Priority, Lane). Without ProjectV2, it has no data source and hard-errors.

Acceptance:
- Remove `dashboard.ts` or replace with a plain `gh issue list` view
- Drop `fetchAllItems()` hard dependency on `GH_PROJECT_ID`
- Clean up HTML rendering logic tied to `fieldValues` nodes

Related: discussion on dropping ProjectV2 in favor of plain GitHub issues.

Issue #240, opened 2026-05-31
Reconcile 4 drifted/divergent dev-init↔dev-core shared/ files


MickaelV0:
**Origin:** #232 (deferred follow-up — file reconciliation, split out of #232 scope)

## Context

#232 evaluated all `dev-init↔dev-core shared/` pairs. Four files were assigned governance value `leave-alone-pending-reconciliation` in #232's ADR because dev-init has drifted behind or forked from dev-core. They require reconciliation before they can join the copy-sync manifest or be documented as intentional forks.

## 4 Files to Reconcile

| File | Status | Issue |
|------|--------|-------|
| `adapters/github-infra.ts` | dev-init behind | dev-init throws raw `Error` instead of `ConfigError` — latent bug |
| `domain/parse-issue-ref.ts` | dev-init behind | dev-init missing `formatRef()` export |
| `ports/workspace.ts` | dev-init behind | dev-init missing `vercelProjectId`/`vercelTeamId` fields |
| `__tests__/domain.test.ts` | divergent | non-trivial divergence, needs analysis |

## Task

For each file:
1. Reconcile dev-init → dev-core (bring dev-init up to date)
2. Then: either add to the copy-sync manifest (`tools/shared-sources.json`) if fully reconciled, OR document as intentional fork if the divergence is intentional

## Out of Scope

`adapters/github-adapter.ts` is already recorded as `intentional-fork` (+272 lines). Do **not** include it as a reconcile target here.

## Notes

- Pure hygiene, not urgent.
- Unblocked once #232 is merged (ADR and copy-sync baseline established).

PR #263, opened 2026-06-08
feat(issues): switch read path to repo-centric (drop ProjectV2)

### LINKED ISSUE #253: Switch CLI issues to repo-centric (drop ProjectV2)
`cli/commands/issues.ts` currently fetches issues via `ProjectV2.items` GraphQL queries. It should be switched to a repo-centric model using `gh issue list` or `repository.issues` GraphQL.

Acceptance:
- Replace `fetchProjectItems(projectId)` with `fetchRepoIssues(repo)`
- Map labels (`status:`, `size:`, `priority:`, `lane:`) to `RawItem` shape instead of `fieldValues`
- Keep `-A` flag as a loop over workspace repos
- Drop `projectId` from `WorkspaceProject` or make it optional
- Update `__tests__/issues.test.ts` fixtures

Effort estimate: ~200 lines, 1 core file.

MickaelV0:
## Summary
- Replace the ProjectV2-board read path in the `roxabi issues` CLI with a direct `repository.issues` GraphQL fetch. Under the issues-only model (#260) status/size/priority/lane live as labels, so the board (and per-repo `projectId`) is redundant for reading.
- `RawItem.fieldValues` is **synthesized from labels** (reverse-lookup of the `*_LABEL_MAP` tables) so every formatter (`formatTable`/`formatTree`/`formatJson`) keeps working with zero change. Priority is a map-key reverse-lookup (`P1-high` → `P1 - High`), not a prefix strip.
- `-A` now loops one fetch per workspace repo (no 100-item batched cap). `fetchProjectItems` + the `ISSUES_QUERY`/`buildBatchedQuery`/`buildBatchedVariables` imports are removed from the read path.

This is the foundational repo-centric shape that #252 (drop ProjectV2 dashboard) will consume.

## Governance (zero edits to held/synced files)
- `shared/types.ts` (copy-synced), `shared/queries.ts` (copy-synced), `shared/ports/workspace.ts` (#240 hold) — **untouched** (verified `git diff` empty).
- `REPO_ISSUES_QUERY` is kept **local to `issues.ts`** (not added to the copy-synced shared query module).
- `projectId` stays required on `WorkspaceProject`; the read path simply stops reading it (optionalization deferred to #240/#252).

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #253: Switch CLI issues to repo-centric (drop ProjectV2) | OPEN |
| Frame | [253-…-frame.mdx](artifacts/frames/253-switch-cli-repo-centric-frame.mdx) | Present |
| Spec | [253-…-spec.mdx](artifacts/specs/253-switch-cli-repo-centric-spec.mdx) | Present |
| Plan | [253-…-plan.mdx](artifacts/plans/253-switch-cli-repo-centric-plan.mdx) | Present |
| Implementation | 1 impl commit on `feat/253-switch-cli-repo-centric` | Complete |
| Verification | Lint ✅ Typecheck ✅ Tests ✅ (full suite 857 pass / 0 fail) | Passed |

## Test Plan
- [x] Unit: `labelsToFieldValues` maps `status:`/`size:`/`P[0-3]-`/`graph:lane/` → canonical field values; unknown labels dropped; **`P1-high` → `P1 - High`** reverse-lookup guard.
- [x] Golden parity: a label-synthesized `RawItem` renders byte-identical to a hand-built `fieldValues`-based `RawItem` via `formatTable`.
- [x] Integration: `-A` fires N requests (one per repo); grouping under `## <label>`; default cwd path resolves a single repo.
- [x] SC7: a workspace project with **no `projectId`** still returns its open issues (read path never touches `projectId`).

Fixes #253

---

### DIFF
--- a/artifacts/frames/253-switch-cli-repo-centric-frame.mdx
+++ b/artifacts/frames/253-switch-cli-repo-centric-frame.mdx
@@ -1,0 +1,82 @@
+---
+title: Switch CLI issues to repo-centric (drop ProjectV2)
+issue: 253
+status: approved
+tier: F-lite
+date: 2026-06-08
+---
+
+## Problem
+
+`roxabi issues` (`cli/commands/issues.ts`) reads issues by querying `ProjectV2.items`
+GraphQL — single-project via `fetchProjectItems(projectId)` and `-A` via a batched
+`buildBatchedQuery(projectIds)`. Both require every workspace repo to be attached to a
+GitHub ProjectV2 board and carry a `projectId`. The board is now redundant under the
+issues-only model adopted in #260 (status/lane/size/priority live as labels, not board
+fields). A repo with no project board returns nothing from the current read path even
+though its issues exist. The CLI must read issues directly from the repository.
+
+This is the foundational shape change that #252 (drop ProjectV2 dashboard) consumes:
+#253 defines the repo-centric fetch + label→`RawItem` mapping; #252 conforms to it.
+
+## Who
+
+- **Primary:** maintainers running `roxabi issues` / `roxabi issues -A` to triage across
+  workspace repos — today blocked on board membership + `projectId` per repo.
+- **Secondary:** #252 (dashboard) which will build on the same repo-centric fetch shape.
+
+## Constraints
+
+- **Shared-file governance — zero edits to copy-synced / held files:**
+  - `shared/types.ts` (`RawItem`) is **copy-synced** to dev-init (CI byte-equality gate).
+    → Keep `RawItem` unchanged. Build issues into the existing shape.
+  - `shared/queries.ts` (`ISSUES_QUERY`, `buildBatchedQuery`) is **copy-synced**.
+    → Keep the new repo-centric GraphQL query **local to `issues.ts`**; do not add it to
+    the shared module. Old project queries stay (still used by #252 dashboard until it lands).
+  - `shared/ports/workspace.ts` (`WorkspaceProject.projectId`) is on the **#240
+    leave-alone-pending-reconciliation hold** — ¬touch.
+- **Formatters read `item.fieldValues.nodes`** (`table-formatter.ts::fieldValue(item,'Status'|'Size'|'Priority')`).
+  → Synthesize `fieldValues.nodes` from `status:`/`size:`/`priority:`/`lane:` labels so the
+  table/tree/json formatters keep working with no downstream change.
+- Single domain (CLI/backend TS), bun + vitest + biome.
+- Preserve current UX: `--tree`/`-T`, `--json`, `--all`/`-A`, cwd→project matching.
+
+## Out of Scope
+
+- **Drop/optionalize `projectId` on `WorkspaceProject`** — deferred to #240/#252
+  (workspace.ts is on the #240 reconciliation hold). #253 leaves `projectId` required and
+  simply stops reading it in the issues fetch path.
+- **`dashboard.ts` repo-centric switch** — that is #252.
+- **Editing copy-synced shared files** (`types.ts`, `queries.ts`) — avoided by design.
+- Removing the old project GraphQL queries — they remain until #252 retires the dashboard.
+
+## Premise Validity
+
+**Success in 6 months:** `roxabi issues` and `roxabi issues -A` list open issues for every
+workspace repo with **no ProjectV2 board configured** — issues read directly from the
+repository; the issues read path issues **zero** `ProjectV2.items` GraphQL calls; #252
+builds on the repo-centric fetch without rework.
+
+**Failure in 6 months:** `issues.ts` still issues a `ProjectV2.items` query in its read
+path (grep `ISSUES_QUERY`/`node.items`/`buildBatchedQuery` in `issues.ts` > 0 matches),
+OR `roxabi issues` returns empty/errors against a repo that has no project board. Either
+is a measurable, observable boolean state at the read path.
+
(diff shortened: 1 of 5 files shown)


MickaelV0:
## Code Review — PR #263 (repo-centric issues CLI)

Fresh multi-agent review (security · backend · tester · product-lead). Single chunk, 2 code files. Secret scan clean. QG verified green by lead (lint/typecheck/test all exit 0, 857 pass).

**Verdict: Approve with comments.** No happy-path correctness bug; the feature works and all 8 spec success criteria are met. Three worthwhile robustness/coverage improvements below (none block merge).

### Warnings (worth fixing — cheap, real)

**suggestion(blocking): multi-page pagination is never exercised by a test**
`plugins/dev-core/cli/__tests__/issues.test.ts`
-- tester
Root cause: every fixture hard-codes `{ hasNextPage: false, endCursor: null }`. `fetchRepoIssues`' `do/while(cursor)` loop would still pass all tests if the cursor-advance were broken or removed.
Class: [test-tautology]
Solution: add one test — call 0 returns `{hasNextPage:true,endCursor:'abc'}` + node A; call 1 (cursor `'abc'`) returns `{hasNextPage:false}` + node B; assert both issues render and fetch was called twice for a single-project workspace.
Confidence: 95%

**suggestion(blocking): `repo.split('/')` is unguarded**
`plugins/dev-core/cli/commands/issues.ts:99`
-- backend-dev
Root cause: a stored `repo` without `/` yields `name = undefined`, coerced to the string `"undefined"`, sent to GitHub; `repository` returns `null` and the crash surfaces two lines later at `data.data.repository.issues` rather than at the source.
Class: [missing-input-validation]
Solution: `if (!owner || !name) throw new Error(\`Invalid repo "${repo}" — expected owner/name\`)` before building variables.
Confidence: 90%

**issue: `repository: null` (repo not found / no scope) is a 200 with null data → null-deref**
`plugins/dev-core/cli/commands/issues.ts:105-110`
-- backend-dev
Root cause: GraphQL resource/permission errors arrive as HTTP 200 with `data.repository = null`; `res.ok` is true so `ghGraphQL` returns, and the type cast omits the `null` branch. (Pre-existing pattern carried from the old `node.items` path — not a regression, but now easy to harden.)
Class: [missing-error-handling]
Solution: widen the cast to `repository: {…} | null` and throw a clear "Repository not found or inaccessible" before reading `.issues`; optionally surface a top-level `errors[]`.
Confidence: 85%

- nitpick: `runIssuesCommand` re-filters `i.content?.state === 'OPEN'` though `REPO_ISSUES_QUERY` already passes `states:[OPEN]` — redundant for top-level items (harmless; can drop). -- product-lead
- suggestion(non-blocking): the `-A` loop fetches repos sequentially (N round-trips). Intentional (named in PR) to drop the 100-item batched cap; `Promise.all(ws.projects.map(...))` would parallelize if latency matters at scale. -- backend-dev
- suggestion(non-blocking): `buildBatchedQuery`/`buildBatchedVariables` describe blocks remain in `issues.test.ts`. **Deliberate plan decision** — those functions still exist in `queries.ts` for #252 and the tests still pass; they are orthogonal to `issues.ts`. Consider relocating to a `queries.test.ts` so they don't read as issues-command coverage. -- tester / product-lead
- thought: per-repo request tests assert count + endpoint but not the request *body* (`variables.owner/repo`); a hardcoded-repo regression wouldn't be caught. Add `JSON.parse(calls[i][1].body).variables` assertions if you want routing locked. -- tester
- thought: module-scope reverse index silently overwrites on label collision (disjoint today). A load-time `if (idx.has(name)) throw` would catch future regressions. -- backend-dev

### Praise
- No injection: `owner`/`name`/`cursor` pass as GraphQL **variables**, never interpolated into the static query; Bearer token never logged. -- security-auditor
- Reverse-lookup guard (`P1-high → 'P1 - High'`) and golden-parity test are correctly structured and non-tautological. -- tester
- All 8 success criteria met; `REPO_ISSUES_QUERY` kept local; zero diff in the copy-synced / #240-held shared files. -- product-lead

MickaelV0:
## Review Fixes Applied

**Applied via 1b1:** 3 finding(s) (commit `9036b2a`)
**Deferred:** 0 · **Skipped (non-blocking):** 5 · **Failed:** 0

### Applied
- [applied] suggestion(blocking): multi-page pagination now exercised — new integration test asserts the `do/while` loop fires a 2nd request, threads `endCursor`→`after`, and accumulates both pages. `issues.test.ts` (tester)
- [applied] suggestion(blocking): `repo.split('/')` guarded — throws "expected owner/name" instead of a late null-deref. `issues.ts:100` (backend-dev)
- [applied] issue: `data.repository === null` (repo not found / no scope, HTTP 200) now throws "not found or inaccessible" before reading `.issues`. `issues.ts:105-112` (backend-dev)

### Skipped (non-blocking — left as-is by design)
- redundant top-level `state === 'OPEN'` filter (query already `states:[OPEN]`) — harmless
- sequential `-A` fetch — intentional (drops 100-item batched cap)
- `buildBatchedQuery`/`buildBatchedVariables` tests retained in this file — deliberate plan decision (still exported for #252; still pass)
- per-repo request-body assertions, reverse-index collision assert — nice-to-haves, not regressions

QG re-verified green by lead: lint 0 warnings/errors · typecheck 0 · test 0 failures (`issues.test.ts` 21 pass). Labeled `reviewed` for squash auto-merge on green CI.

PR #295, opened 2026-06-11
feat(dev-core): executed falsification gate — prove tests fail when feature breaks (#280)

MickaelV0:
## Summary

- Upgrades the Negative-Test Rule from a mental check to an executed falsification gate across 3 files
- `implement/SKILL.md`: adds Step 6b (gate runs immediately after SC→Test Matrix; stash → FAIL → pop → green → evidence)
- `test/SKILL.md`: adds Step 8 (standalone `/test` gate; owned by calling agent; e2e exempt)
- `agents/tester.md`: replaces "mentally deleting the guard" with the mandatory executed protocol; unproven test = `⏳ not run`, tautological = merge blocker
- `plugin.json`: version bump 0.8.1 → 0.8.2 (triggered by check-skill-version hook)

Closes #280

## SC→Test Matrix

| SC | Test file | Test name | Status |
|----|-----------|-----------|--------|
| implement Step 6b: stash source → assert FAIL | — | — | ⚠ NO TEST — prompt-logic-only |
| implement Step 6b: pop stash → assert GREEN | — | — | ⚠ NO TEST — prompt-logic-only |
| implement Step 6b: tautological → blocking gap reported | — | — | ⚠ NO TEST — prompt-logic-only |
| implement Step 6b: evidence line appended to summary | — | — | ⚠ NO TEST — prompt-logic-only |
| implement Step 6b: e2e rows annotated NO FALSIFY | — | — | ⚠ NO TEST — prompt-logic-only |
| test Step 8: gate owned by calling agent (¬tester) | — | — | ⚠ NO TEST — prompt-logic-only |
| test Step 8: e2e → annotate NO FALSIFY, stop | — | — | ⚠ NO TEST — prompt-logic-only |
| test Step 8: tautological → DP(A) Fix \| Flag, ¬silently pass | — | — | ⚠ NO TEST — prompt-logic-only |
| tester.md: ¬mental-only — evidence line required | — | — | ⚠ NO TEST — prompt-logic-only |
| tester.md: no evidence line → Status ⏳ not run | — | — | ⚠ NO TEST — prompt-logic-only |
| tester.md: tautological → merge blocker, ¬✓ proven | — | — | ⚠ NO TEST — prompt-logic-only |

All rows `⚠ NO TEST — prompt-logic-only`: gate logic lives in LLM-executed SKILL.md instructions; no executable test harness can assert that an agent follows a process step.

## Test plan

- [x] `bun run test` — 492 tests passing (33 files)
[...]

Issue #288, opened 2026-06-11
checkup: conditional-job skip-rate detector (dead-gate per-job dormancy, #579 class)


MickaelV0:
## Origin

Split out of #278 (dead-gate detector, action P1-5). #278 delivers the **token** sub-check + **push-gate merge-relative** sub-check; the third sub-check — per-job dormancy — was descoped because it needs job-level YAML parsing + per-job run history, disproportionate to the other two. This is that follow-up. Same root class as #579 (e2e suite that existed but never ran).

## Problem

A gate can be alive at the *workflow* level (the file runs on push, the token is safe) yet a specific **job** inside it never executes: a path filter never matches, an `if:` condition is permanently false, a matrix leg is empty, a required-check job is silently skipped. #278 sees the workflow run and reports the gate healthy; the dormant job is invisible. #579's e2e suite was exactly this — the workflow ran, the e2e job was filtered out, and 4 regressions shipped under a green check.

## Mechanism

Extends `checkDeadGates()` in `dev-core:checkup` (`doctor-github.ts`) with a **per-job** dormancy pass. Read-only.

1. **Parse jobs:** for each push-/PR-triggered workflow, enumerate jobs + their `if:`, `paths:`/`paths-ignore:`, and matrix expansion (static YAML — no execution).
2. **Per-job run history:** cross-reference `gh run list` job-level data (`gh run view <id> --json jobs`) over the observation window → for each job, count executions vs the number of runs where it *could* have run.
3. **Skip-rate verdict:** a job that was eligible (workflow ran) but executed **0 times across ≥N runs** → flag as dormant. A job skipped *every* time its workflow ran is either dead wiring or a permanently-false guard.
4. **Required-check cross-check:** if a dormant job is also a **required status check** (branch protection), escalate the finding — a required check that never runs is a merge gate that proves nothing (the #579 signature).
5. **Grace window + anchor:** same merge-relative anchoring as #278 — don't flag a young job or a genuinely-rare path.

## Where

`plugins/dev-core/skills/checkup/doctor-github.ts` (extend `checkDeadGates()` with a `detectDormantJobs()` pass — lands AFTER #278's `checkDeadGates()` + `detectUnsafeTokenInTriggeredWorkflow()` exist) + `plugins/dev-core/skills/checkup/__tests__/dead-gate.test.ts` (job-dormancy cases). Reuses #278's `STANDARD_WORKFLOWS` iteration + grace-window helper — **do not** duplicate them.

[...]

Issue #280, opened 2026-06-11
test: falsification gate — prove new tests fail when the feature is broken


MickaelV0:
## Origin

Daily review 2026-06-10 — Lot 3, action **P1-7**. The boilerplate e2e suite (#579) was green-looking while structurally broken: tests that cannot fail pass forever and prove nothing.

## Problem

Always-green tests — wrong selector, wrong env wiring, silent skip, assertion mocked away — are worse than no tests: they convert "unproven" into "proven" in every report. Nothing in the current test phase distinguishes a test that passes *because the feature works* from one that passes *no matter what*.

## Prior art (already in the tester agent)

`plugins/dev-core/agents/tester.md` already carries the **"Negative-Test Rule (MANDATORY — merge blocker)"** + "Tautological test signals" — but as a *mental* check ("Verify by **mentally** deleting the guard"). This issue upgrades it from mental to **executed**.

## Mechanism

For each new/modified test written as SC proof, in the **`implement` Verify phase** (the agent that owns source), or standalone `/test`:

1. **Falsify via `git stash`:** stash the implementation (source only), run the test, **assert it FAILS**. (`git stash`, not commit-revert → no detached HEAD; not targeted mutation → reliable restore.)
2. **`git stash pop`**, assert green again.
3. **Record evidence:** `broke <X> → test failed with <Y>` — one line per test, feeding the #279 matrix `Status` column.
4. **Boundary respected:** the tester writes test files only (`agents/tester.md` boundary "¬source code"); the falsification (a *source* mutation) is driven by the **implement orchestrator / domain agent** that owns source, via stash. The tester's Negative-Test Rule is updated to *require* the executed evidence line, not run the stash itself.
5. **Scope:** unit + fast-integration tests only. e2e tests are **exempt** via an explicit `NO FALSIFY — e2e` annotation (full stash+rerun is too expensive / context-bound). Not retroactive to existing suites (the dead-gate detector #278 covers the legacy side).

## Where

`plugins/dev-core/skills/implement/SKILL.md` (Verify step — the executed falsification, owns source) + `plugins/dev-core/skills/test/SKILL.md` (falsification policy for standalone `/test`) + `plugins/dev-core/agents/tester.md` (Negative-Test Rule: mental → require executed evidence; tester stays ¬source).

## Trigger

Every test/Verify phase of an implementation (`/test`, and the Verify step inside `/dev`).

## Acceptance criteria

[...]

PR #296, opened 2026-06-11
feat(checkup): dead-gate detector — flag CI gates that never run (#278)

### LINKED ISSUE #278: checkup: dead-gate detector — flag CI gates that never actually run
## Origin

Daily review 2026-06-10 — Lot 3, action **P1-5**. The class was proven live twice in the same week on roxabi-boilerplate:

- **#579 (e2e never-run):** the e2e suite existed and looked healthy for months but never actually executed (path filter + required-check wiring). First real run unmasked **4 shipped regressions** (vite8+nitro3 broken SSR bundle, import-type DI break, …).
- **#589 (push CI never triggered):** auto-merge was armed with `github.token`; GitHub anti-recursion means pushes attributed to that token never trigger `push` workflows → **every merge to staging ran zero post-merge CI** since the gate existed.

In both cases everything *looked* green: gates passed on PRs. The failure mode is "never runs where it matters" — invisible in day-to-day signals.

## Problem

A gate that exists, is green on PRs, but never executes post-merge (or never executes at all) is indistinguishable from a healthy gate. Discovery today is accidental archaeology, months later, usually triggered by a shipped regression.

## Mechanism

New check `checkDeadGates()` in `dev-core:checkup` (`doctor-github.ts`), per repo. Read-only.

1. **Enumerate** workflows + trigger events (extends the existing `checkWorkflows()` which today only checks file presence).
2. **Push-gate history (merge-relative — NOT a bare count):** for each push-triggered workflow on staging/main, flag if **0 push-event runs landed while ≥N merge-commits landed** on that branch (N default 5). Cross-reference `gh run list --branch X --event push` with the branch's merge count. Anchored to real activity, so a healthy slow-cadence repo isn't flagged and an ancient-but-dead gate is. (A bare "0 of last 5 runs" can span 18 months and lie — rejected.)
3. **Token check (3-state taxonomy — post roxabi-ci App migration, see #284 pilot + #289 generator-side App-mode):** static YAML scan of push-/bot-triggered steps:
   | Token | Re-triggers CI? | Verdict |
   |---|---|---|
   | `github.token` / `secrets.GITHUB_TOKEN` | ✗ | **DEAD GATE — flag** (#589 class) |
   | `secrets.PAT` | ✓ | valid (legacy); optional `warn` "migrate to roxabi-ci App" |
   | `actions/create-github-app-token` → `steps.<id>.outputs.token` | ✓ | **valid, canonical — must NOT false-positive** |
   During coexistence the fleet is mixed PAT/App → accept **both** without false positives. Cheapest sub-check; pure static.
4. **Grace window:** skip workflows/repos younger than the observation window (`gh repo view --json createdAt`).
5. **Output:** one finding per dead gate in a **`Dead Gates`** doctor Section, with evidence (last real push-run date, merges-without-run count).

## Out of scope → follow-up issue

**Conditional-job skip-rate detector** (per-job `if:`/path-filter dormancy — the #579 class) is split out: it needs job-level YAML parsing + per-job run history, disproportionate to the token + push-gate sub-checks. Tracked as **#288** (blocked-by this), not delivered here.

## Where

`plugins/dev-core/skills/checkup/doctor-github.ts` (new `checkDeadGates()` + `detectUnsafeTokenInTriggeredWorkflow()` static analyzer, following the `detectMissingContentsRead` pattern) + `plugins/dev-core/skills/checkup/__tests__/dead-gate.test.ts` + one import/array line in `doctor.ts`. Iterate over `STANDARD_WORKFLOWS` (`shared/adapters/github-infra.ts`), never hard-coded filenames. `cookbooks/devcore-checks.md` gets an editorial row only (no logic lives there).

## Trigger

Every `/checkup` run. Read-only → cron-safe.

## Acceptance criteria

- [ ] `github.token`/`secrets.GITHUB_TOKEN` in a push-triggered step is flagged; `secrets.PAT` and `steps.*.outputs.token` (App) are NOT flagged (no false positive in a mixed PAT/App fleet).
- [ ] A push-triggered workflow with 0 push-event runs while ≥N merges landed on staging is reported (roxabi-boilerplate pre-#589 state would have flagged it).
- [ ] Grace window: workflows younger than the window are skipped.
- [ ] Read-only: never mutates repo or run state.
- [ ] Findings render in a `Dead Gates` Section in `doctor` output.

MickaelV0:
## Summary

Adds a **dead-gate detector** to `dev-core:checkup` — surfaces CI gates that exist and look green on PRs but never actually run where it matters (the #579/#589 class). Read-only.

Two sub-checks (per-job dormancy is split out to **#288**, blocked-by this):

1. **Token taxonomy** — `detectUnsafeTokenInTriggeredWorkflow()` static-YAML scan of push-/bot-triggered steps:
   - `github.token` / `secrets.GITHUB_TOKEN` → **DEAD GATE** (anti-recursion: pushes never re-trigger CI)
   - `secrets.PAT` → **warn** (valid legacy, migrate to roxabi-ci App)
   - `actions/create-github-app-token` → `steps.*.outputs.token` → **canonical-valid**, never flagged
2. **Push-gate merge-relative history** — flags a push-triggered workflow with **0 push-event runs while ≥N merge-commits landed** (N=5 default), anchored to real branch activity (not a bare "0 of last 5").

Iterates `STANDARD_WORKFLOWS` (not `REQUIRED_SECRETS` keys) so `ci.yml`/`pr-title.yml` are covered. Renders a `Dead Gates` section in `doctor` output.

Closes #278

## SC → Test matrix

| SC | Test(s) | Status |
|---|---|---|
| SC1: `github.token`/`GITHUB_TOKEN` flagged; `secrets.PAT` & App `outputs.token` NOT flagged (no FP in mixed fleet) | `detectUnsafeTokenInTriggeredWorkflow :: AC-1 (env push / GITHUB_TOKEN / on:[push] / workflow_run)`, `AC-2 :: secrets.PAT as pat-warn`, `AC-3 :: does not flag steps.*.outputs.token (App)` + `checkDeadGates :: AC-3 App token not flagged` | ✓ proven |
| SC2: push-triggered workflow with 0 push-runs while ≥N merges landed is reported | `checkDeadGates (subprocess) :: AC-4 + AC-5: flags push-gate dead gate in JSON output` | ✓ proven |
| SC3: grace window — workflows younger than the window are skipped | grace-window guard implemented in `checkDeadGates` via repo `createdAt`; no dedicated unit test | ⚠ NO TEST — infra-not-wired |
| SC4: read-only — never mutates repo or run state | `checkDeadGates (subprocess) :: AC-1 (read-only): doctor does not mutate repo or run state` | ✓ proven |
| SC5: findings render in a `Dead Gates` Section | `checkDeadGates (subprocess) :: AC-5: Dead Gates section appears in output` | ✓ proven |

14/14 tests in `dead-gate.test.ts` green · full suite 506 passed / 4 skipped · typecheck clean · biome clean.

## Notes

- **Recovered run:** the first implementation agent completed the code + tests but hit a context-overflow before committing; a second agent fixed the 6 then-failing token-scanner assertions (the static scanner wasn't matching tokens); the lead finalized gates + commit + PR.
- **Version bump coordination:** bumps `dev-core` `0.8.1 → 0.8.2`, same as #295 (#280). Identical 3-way change → clean auto-merge; whichever of #278/#295 lands second needs no manual rebase on the version line.
- Does **not** add `detectDormantJobs()` (per-job dormancy) — that is #288, which extends `checkDeadGates()`.

### DIFF
--- a/plugins/dev-core/skills/checkup/__tests__/dead-gate.test.ts
+++ b/plugins/dev-core/skills/checkup/__tests__/dead-gate.test.ts
@@ -296,7 +296,7 @@ describe('checkDeadGates (integration via subprocess)', () => {
         '  "api repos/TestOrg/test-repo/branches/staging") echo "{}"; exit 0 ;;',
         '  "api repos/TestOrg/test-repo/branches/main") echo "{}"; exit 0 ;;',
         // push run count: return 0 (no push runs landed)
-        '  "run list"*"--event" "push"*"--json" "databaseId"*) echo "0"; exit 0 ;;',
+        '  "run list"*"--event"*"push"*"--json"*"databaseId"*) echo "0"; exit 0 ;;',
         // merge commit count: ≥5 merges
         '  "api repos/TestOrg/test-repo/commits"*) echo "7"; exit 0 ;;',
         '  "label list"*) echo "[]"; exit 0 ;;',
@@ -322,8 +322,10 @@ describe('checkDeadGates (integration via subprocess)', () => {
     expect(result.stdout).toContain('Dead Gates')
   })
 
-  it('AC-4 + AC-5: flags push-gate dead gate in JSON output', () => {
-    // Write a push-triggered workflow with unsafe token
+  it('AC-4: flags push-gate dead gate in JSON output (push-gate history sentinel)', () => {
+    // Write a push-triggered workflow using a safe App token so that token taxonomy passes —
+    // only the push-gate history sub-check (0 push runs + ≥5 merges) can produce a fail.
+    // This isolates AC-4: a regression in push-gate logic stays visible even if token taxonomy passes.
     const wfDir = path.join(tmpDir, '.github', 'workflows')
     fs.mkdirSync(wfDir, { recursive: true })
     fs.writeFileSync(
@@ -337,21 +339,32 @@ describe('checkDeadGates (integration via subprocess)', () => {
         '  build:',
         '    runs-on: ubuntu-latest',
         '    steps:',
+        '      - name: Mint app token',
+        '        id: app',
+        '        uses: actions/create-github-app-token@v1',
+        '        with:',
+        '          app-id: ${{ vars.APP_ID }}',
+        '          private-key: ${{ secrets.APP_PRIVATE_KEY }}',
         '      - name: Push',
         '        run: git push',
         '        env:',
-        '          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}',
+        '          GH_TOKEN: ${{ steps.app.outputs.token }}',
       ].join('\n'),
     )
 
     const result = runDoctor(tmpDir, ['--json'])
-    const parsed = JSON.parse(result.stdout) as Array<{ name: string; checks: Array<{ status: string }> }>
+    const parsed = JSON.parse(result.stdout) as Array<{
+      name: string
+      checks: Array<{ name?: string; status: string; detail?: string }>
+    }>
 
     const deadGatesSection = parsed.find((s) => s.name === 'Dead Gates')
     expect(deadGatesSection).toBeDefined()
-    // Should have at least one fail (token taxonomy)
-    const hasFail = deadGatesSection?.checks.some((c) => c.status === 'fail')
-    expect(hasFail).toBe(true)
+    // Must have a push-gate history fail with the canonical sentinel phrase
+    const pushGateFail = deadGatesSection?.checks.find(
+      (c) => c.status === 'fail' && c.detail?.includes('never actually runs on pushes'),
+    )
+    expect(pushGateFail).toBeDefined()
   })
 
   it('AC-3: does not flag App token in push-triggered workflow', () => {
@@ -384,15 +397,17 @@ describe('checkDeadGates (integration via subprocess)', () => {
     const result = runDoctor(tmpDir, ['--json'])
     const parsed = JSON.parse(result.stdout) as Array<{
       name: string
-      checks: Array<{ status: string; detail?: string }>
+      checks: Array<{ name?: string; status: string; detail?: string }>
     }>
 
     const deadGatesSection = parsed.find((s) => s.name === 'Dead Gates')
     expect(deadGatesSection).toBeDefined()
 
-    // Token taxonomy check must NOT contain a 'dead' fail about this workflow
-    const tokenFails = deadGatesSection?.checks.filter((c) => c.status === 'fail' && c.detail?.includes('DEAD GATE'))
-    expect(tokenFails).toHaveLength(0)
+    // Token taxonomy check (name === 'token taxonomy') must NOT flag this workflow as a dead gate.
+    // Scope to the taxonomy check only — push-gate history checks use the same phrase but are separate.
+    const tokenTaxonomyCheck = deadGatesSection?.checks.find((c) => c.name === 'token taxonomy')
+    expect(tokenTaxonomyCheck).toBeDefined()
+    expect(tokenTaxonomyCheck?.status).not.toBe('fail')
   })
 
   it('AC-1 (read-only): doctor does not mutate repo or run state', () => {
--- a/plugins/dev-core/skills/checkup/__tests__/dead-gate.test.ts
+++ b/plugins/dev-core/skills/checkup/__tests__/dead-gate.test.ts
@@ -191,6 +191,64 @@ describe('detectUnsafeTokenInTriggeredWorkflow', () => {
     })
   })
 
+  describe('AC-3b (review #296): job-level env + no safe-token masking', () => {
+    it('flags github.token declared at JOB-LEVEL env (not in a step)', () => {
+      const content = [
+        'name: Release',
+        'on:',
+        '  push:',
+        '    branches: [main]',
+        'jobs:',
+        '  publish:',
+        '    runs-on: ubuntu-latest',
+        '    env:',
+        '      GH_TOKEN: ${{ github.token }}',
+        '    steps:',
+        '      - run: gh release create',
+      ].join('\n')
+      const issues = detectUnsafeTokenInTriggeredWorkflow(content, 'release.yml')
+      const dead = issues.filter((i) => i.kind === 'dead')
+      expect(dead).toHaveLength(1)
+      expect(dead[0].step).toBe('(job-level env)')
+    })
+
+    it('flags a dead token even when a safe App token is present in the same step (no masking)', () => {
+      const content = [
+        'name: Auto Merge',
+        'on:',
+        '  push:',
+        '    branches: [staging]',
+        'jobs:',
+        '  merge:',
+        '    runs-on: ubuntu-latest',
+        '    steps:',
+        '      - name: Mixed',
+        '        run: gh pr merge',
+        '        env:',
+        '          APP: ${{ steps.app.outputs.token }}',
+        '          GH_TOKEN: ${{ github.token }}',
+      ].join('\n')
+      const issues = detectUnsafeTokenInTriggeredWorkflow(content, 'auto-merge.yml')
+      expect(issues.filter((i) => i.kind === 'dead')).toHaveLength(1)
+    })
+
+    it('flags secrets.PAT at job-level env as pat-warn', () => {
+      const content = [
+        'name: Sync',
+        'on: push',
+        'jobs:',
+        '  sync:',
+        '    runs-on: ubuntu-latest',
+        '    env:',
+        '      GH_TOKEN: ${{ secrets.PAT }}',
+        '    steps:',
+        '      - run: git push',
+      ].join('\n')
+      const issues = detectUnsafeTokenInTriggeredWorkflow(content, 'sync.yml')
+      expect(issues.filter((i) => i.kind === 'pat-warn')).toHaveLength(1)
+    })
+  })
+
   describe('edge cases', () => {
     it('does not flag push-triggered workflows with no token usage at all', () => {
       const content = [
--- a/plugins/dev-core/skills/checkup/__tests__/dead-gate.test.ts
+++ b/plugins/dev-core/skills/checkup/__tests__/dead-gate.test.ts
@@ -1,0 +1,410 @@
+import * as fs from 'node:fs'
+import * as os from 'node:os'
+import * as path from 'node:path'
+import { afterEach, beforeEach, describe, expect, it } from 'vitest'
+import { detectUnsafeTokenInTriggeredWorkflow } from '../doctor-github'
+
+/**
+ * Dead Gate detector — acceptance tests.
+ *
+ * AC-1: github.token / secrets.GITHUB_TOKEN in push-triggered steps → flagged as 'dead'
+ * AC-2: secrets.PAT in push-triggered steps → flagged as 'pat-warn' (not 'dead')
+ * AC-3: App token via actions/create-github-app-token → steps.<id>.outputs.token → NOT flagged
+ * AC-4: Push-gate merge-relative history: 0 push runs + ≥N merges → fail (integration via subprocess)
+ * AC-5: findings appear in 'Dead Gates' Section
+ */
+
+// ── Unit tests: detectUnsafeTokenInTriggeredWorkflow ──────────────────────────
+
+describe('detectUnsafeTokenInTriggeredWorkflow', () => {
+  describe('AC-1: github.token / GITHUB_TOKEN → flagged as dead', () => {
+    it('flags github.token in env of push-triggered step', () => {
+      const content = [
+        'name: Auto Merge',
+        'on:',
+        '  push:',
+        '    branches: [staging]',
+        'jobs:',
+        '  merge:',
+        '    runs-on: ubuntu-latest',
+        '    steps:',
+        '      - name: Trigger merge',
+        '        uses: some/action@v1',
+        '        with:',
+        '          token: ${{ github.token }}',
+      ].join('\n')
+
+      const issues = detectUnsafeTokenInTriggeredWorkflow(content, 'auto-merge.yml')
+
+      expect(issues).toHaveLength(1)
+      expect(issues[0].kind).toBe('dead')
+      expect(issues[0].token).toContain('github.token')
+      expect(issues[0].file).toBe('auto-merge.yml')
+      expect(issues[0].job).toBe('merge')
+    })
+
+    it('flags secrets.GITHUB_TOKEN in env of push-triggered step', () => {
+      const content = [
+        'name: CI',
+        'on:',
+        '  push:',
+        '    branches: [main]',
+        'jobs:',
+        '  build:',
+        '    runs-on: ubuntu-latest',
+        '    steps:',
+        '      - name: Push changes',
+        '        run: git push',
+        '        env:',
+        '          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}',
+      ].join('\n')
+
+      const issues = detectUnsafeTokenInTriggeredWorkflow(content, 'ci.yml')
+
+      expect(issues).toHaveLength(1)
+      expect(issues[0].kind).toBe('dead')
+      expect(issues[0].token).toContain('secrets.GITHUB_TOKEN')
+    })
+
+    it('flags github.token in compact on: [push] form', () => {
+      const content = [
+        'name: CI',
+        'on: [push, pull_request]',
+        'jobs:',
+        '  build:',
+        '    runs-on: ubuntu-latest',
+        '    steps:',
+        '      - name: Deploy',
+        '        run: deploy.sh',
+        '        env:',
+        '          TOKEN: ${{ github.token }}',
+      ].join('\n')
+
+      const issues = detectUnsafeTokenInTriggeredWorkflow(content, 'ci.yml')
+
+      expect(issues.some((i) => i.kind === 'dead')).toBe(true)
+    })
+
+    it('flags github.token in workflow_run-triggered step (bot trigger)', () => {
+      const content = [
+        'name: Bot Workflow',
+        'on:',
+        '  workflow_run:',
+        '    workflows: [CI]',
+        '    types: [completed]',
+        'jobs:',
+        '  bot:',
+        '    runs-on: ubuntu-latest',
+        '    steps:',
+        '      - name: Comment',
+        '        uses: some-action@v1',
+        '        with:',
+        '          token: ${{ github.token }}',
+      ].join('\n')
+
+      const issues = detectUnsafeTokenInTriggeredWorkflow(content, 'bot.yml')
+
+      expect(issues.some((i) => i.kind === 'dead')).toBe(true)
+    })
+  })
+
+  describe('AC-2: secrets.PAT → warn (not dead)', () => {
+    it('flags secrets.PAT as pat-warn, not dead', () => {
+      const content = [
+        'name: Auto Merge',
+        'on:',
+        '  push:',
+        '    branches: [staging]',
+        'jobs:',
+        '  merge:',
+        '    runs-on: ubuntu-latest',
+        '    steps:',
+        '      - name: Trigger',
+        '        uses: some/action@v1',
+        '        with:',
+        '          token: ${{ secrets.PAT }}',
+      ].join('\n')
+
+      const issues = detectUnsafeTokenInTriggeredWorkflow(content, 'auto-merge.yml')
+
+      expect(issues).toHaveLength(1)
+      expect(issues[0].kind).toBe('pat-warn')
+      // Must NOT be classified as dead
+      expect(issues.some((i) => i.kind === 'dead')).toBe(false)
+    })
+  })
+
+  describe('AC-3: App token via actions/create-github-app-token → NOT flagged', () => {
+    it('does not flag steps.<id>.outputs.token (App token pattern)', () => {
+      const content = [
+        'name: Auto Merge',
+        'on:',
+        '  push:',
+        '    branches: [staging]',
+        'jobs:',
+        '  merge:',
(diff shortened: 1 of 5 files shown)


plugins/dev-core/skills/checkup/doctor-github.ts
```diff
@@ -486,6 +486,332 @@ export function detectMissingContentsRead(
...
+        // Unnamed step
+        stepHeaders.push({ name: `(unnamed step ${stepHeaders.length + 1})`, start: i })
+      }
+    }
+
+    // Scan each step body for token usage
+    for (let s = 0; s < stepHeaders.length; s++) {
+      const stepStart = stepHeaders[s].start
+      const stepEnd = s + 1 < stepHeaders.length ? stepHeaders[s + 1].start : jobLines.length
+      const stepBody = jobLines.slice(stepStart, stepEnd).join('\n')
```
MickaelV0: **[major · issue] Comment lines not excluded before DEAD_PATTERNS scan → false positive**

The step body is joined into a string and tested against DEAD_PATTERNS without stripping comment lines first. A commented-out token reference like `# token: ${{ github.token }}` inside a step body will trigger a 'dead' finding when no SAFE_PATTERN is present in the same step. Concrete repro:

```yaml
steps:
  - name: Deploy
    run: deploy.sh
    env:
      # GH_TOKEN: ${{ github.token }}  ← old approach, left as comment
      GH_TOKEN: ${{ secrets.MY_PAT }}
```

The DEAD_PATTERN matches the comment, PAT_PATTERN also matches — user gets both a 'dead' and a 'pat-warn' finding on the same step despite the dead token being commented out.

Fix: before joining `stepBody`, filter out lines where `line.trimStart().startsWith('#')`, or apply DEAD_PATTERNS line-by-line with a comment guard. The existing `detectMissingContentsRead` pattern (the model this scanner follows) may have the same gap — worth cross-checking.

<sub>↳ verify: Confirmed genuine defect. At /plugins/dev-core/skills/checkup/doctor-github.ts lines 603-605, `stepBody = jobLines.slice(stepStart, stepEnd).join('\n')` performs a raw join with zero comment filtering. The DEAD_PATTERNS regexes (`/\$\{\{\s*github\.token\s*\}\}/` etc.) are then applied against this raw string. A YAML…</sub>

MickaelV0 (commented):
## 🤖 Automated multi-domain review (Claude, adversarially verified)
Confirmed findings: **major:2 · minor:11 · nit:2 · praise:1**. Inline comments below; unanchored items here.

### 👍 Praise
- **14 unit + integration tests with precise subprocess wiring and clean section naming** — plugins/dev-core/skills/checkup/__tests__/dead-gate.test.ts

### 📌 Unanchored findings (no single diff line)

#### [minor · suggestion] Missing test coverage: commented-out token and job-level env edge cases — `plugins/dev-core/skills/checkup/__tests__/dead-gate.test.ts:—`
The test suite covers the happy-path AC-1 through AC-5 well, but the two realistic mis-classification cases have no tests: (a) a step body containing a YAML comment with `github.token` (no test for false positive suppression), and (b) a job with `env: GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}` at job level with no step-level reference (no test for the false negative). Adding these two cases would pin the current behaviour and make future fixes verifiable.

#### [minor · issue] SC3 grace-window skip path has no unit test — `plugins/dev-core/skills/checkup/__tests__/dead-gate.test.ts:—`
The integration tests all use `createdAt=2020-01-01` (repo age >> 14 days), so they only exercise the grace-window pass path. No test verifies that a repo younger than `GRACE_DAYS=14` days produces a skip/pass result rather than a false positive. The PR acknowledges this gap. Given that SC3 is the boundary most likely to produce silent false negatives on fresh repos, a dedicated unit test is warranted: call `checkDeadGates` (or `repoAgeOk` directly once exported — see companion finding) with `createdAt = new Date().toISOString()` and assert the returned section contains no fails.

#### [minor · suggestion] repoAgeOk is unexported — #288 cannot reuse the grace-window helper without duplication — `plugins/dev-core/skills/checkup/doctor-github.ts:—`
Issue #288 is required to reuse `checkDeadGates`, `STANDARD_WORKFLOWS`, and the grace-window helper for `detectDormantJobs()`. `STANDARD_WORKFLOWS` and `checkDeadGates` are exported, but `repoAgeOk` is a private function. If #288 needs to apply the same 14-day grace window it must either re-implement it or inline it, breaking DRY and risking drift. Exporting `repoAgeOk` now costs nothing and closes the gap: `export function repoAgeOk(createdAt: string, graceDays = GRACE_DAYS): boolean`.

#### [minor · question] workflow_dispatch included as isBotTriggered but no test covers dispatch-only workflows — `plugins/dev-core/skills/checkup/__tests__/dead-gate.test.ts:—`
The unit test suite covers `push`, `pull_request`, `workflow_run`, and combination triggers, but not a workflow whose sole trigger is `workflow_dispatch`. If `detectUnsafeTokenInTriggeredWorkflow` classifies `workflow_dispatch` as a push-equivalent (via `isBotTriggered`), the rationale should be documented and a test added. If `workflow_dispatch` is intentionally out of scope for dead-gate detection, a test asserting it returns `null` / no finding would prevent a future regression that accidentally adds it.

plugins/dev-core/skills/checkup/doctor-github.ts
```diff
@@ -486,6 +486,332 @@ export function detectMissingContentsRead(
...
+
+      // Skip if it uses a safe App token — this step is the mint step itself or consumes app output
+      if (SAFE_PATTERN.test(stepBody)) continue
```
MickaelV0: **[minor · issue] SAFE_PATTERN check is step-scoped but suppresses entire step including a co-present dead token**

If a step body contains both `${{ steps.app.outputs.token }}` (safe) AND `${{ github.token }}` (dead), the SAFE_PATTERN early-return at line 608 silently drops the dead finding. In practice this combination would be unusual (why would one step reference both?), but the code comment says "this step is the mint step itself or consumes app output" — the mint step (`actions/create-github-app-token`) would never also reference `github.token`. However, a misconfigured step that sets up an app token and simultaneously passes `github.token` somewhere else would be missed entirely. Fix: narrow the `continue` guard — only skip the dead-token check if the step uses SAFE_PATTERN *as its token argument* (i.e., the safe token is the one being passed, not just coincidentally present in the step body). This is hard to do without a YAML parser, but at minimum document the known blind spot in a comment.

<sub>↳ verify: Confirmed from the branch code at line 608: `if (SAFE_PATTERN.test(stepBody)) continue` skips the entire dead-token loop whenever any `${{ steps.<x>.outputs.token }}` reference appears anywhere in the step body. SAFE_PATTERN (line 561) and DEAD_PATTERNS (line 559) are independent regex tests on the full stepBody str…</sub>

plugins/dev-core/skills/checkup/doctor-github.ts
```diff
@@ -486,6 +486,332 @@ export function detectMissingContentsRead(
...
+  for (let i = jobsSectionLine + 1; i < lines.length; i++) {
+    const m = lines[i].match(/^ {2}([a-zA-Z0-9][a-zA-Z0-9_-]*):\s*$/)
+    if (m) {
+      jobHeaders.push({ name: m[1], start: i })
+    } else if (/^\S/.test(lines[i]) && lines[i].trim() && !lines[i].trimStart().startsWith('#')) {
+      break
+    }
+  }
+
+  for (let j = 0; j < jobHeaders.length; j++) {
```
MickaelV0, line 583: **[minor · issue] Job-level env: with github.token/secrets.GITHUB_TOKEN is not scanned → false negative**

The scanner only inspects step-level bodies (lines from the step header to the next step header). Job-level `env:` blocks — which inject tokens into all steps of that job — are not scanned. A real-world pattern that is missed:

```yaml
jobs:
  build:
    env:
      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # ← not detected
    steps:
      - run: git push
```

This is a genuine dead-gate (the GITHUB_TOKEN push silently fails), but the scanner reports 'pass'. Fix: after extracting jobLines, do a preliminary scan of lines at 4-space indent that fall before the first step header, applying DEAD_PATTERNS to them. Report findings attributed to the job rather than a specific step (use e.g. `step: '(job-level env)'`).

<sub>↳ verify: Confirmed genuine gap. In the scanner at doctor-github.ts lines 584–637: `jobLines` spans the full job body (line after job header to next job header), but step bodies are sliced starting at `stepHeaders[0].start`. Any job-level `env:` block (4-space indent, before the first `steps:` entry) lives in `jobLines[0 .. s…</sub>

plugins/dev-core/skills/checkup/doctor-github.ts
```diff
@@ -486,6 +486,332 @@ export function detectMissingContentsRead(
...
+
+  // --- 2. Detect steps using unsafe tokens ---
+  // We scan for env vars or `with:` fields containing the dangerous expressions.
+  // Safe: steps.<id>.outputs.token (App token)
+  // Dead: github.token, secrets.GITHUB_TOKEN
+  // Warn: secrets.PAT
+
+  const DEAD_PATTERNS = [/\$\{\{\s*github\.token\s*\}\}/, /\$\{\{\s*secrets\.GITHUB_TOKEN\s*\}\}/]
+  const PAT_PATTERN = /\$\{\{\s*secrets\.PAT\s*\}\}/
+  const SAFE_PATTERN = /\$\{\{\s*steps\.[a-zA-Z0-9_-]+\.outputs\.token\s*\}\}/
```
MickaelV0: **[minor · suggestion] SAFE_PATTERN matches any outputs named 'token', not just app-token outputs**

The pattern `/\$\{\{\s*steps\.[a-zA-Z0-9_-]+\.outputs\.token\s*\}\}/` matches `${{ steps.ANY_ID.outputs.token }}`, where `ANY_ID` can be any step ID — not just an `actions/create-github-app-token` step. A step that produces a `token` output for a different purpose (e.g., a custom action that mints a different kind of token) would suppress the dead-gate check for that step. Consider tightening to: only mark a step as safe when the file also contains `actions/create-github-app-token` as a `uses:` entry (already present in the step body or elsewhere in the job), or add a note in the function doc acknowledging this approximation.

<sub>↳ verify: Confirmed by reading the branch source. At line 521, `SAFE_PATTERN = /\$\{\{\s*steps\.[a-zA-Z0-9_-]+\.outputs\.token\s*\}\}/` matches any step output named `token` regardless of the `uses:` action. The skip logic at line 568 (`if (SAFE_PATTERN.test(stepBody)) continue`) unconditionally suppresses the dead-gate check…</sub>

plugins/dev-core/skills/checkup/doctor-github.ts
```diff
@@ -486,6 +486,332 @@ export function detectMissingContentsRead(
...
+): Array<{ file: string; job: string; step: string; kind: 'dead' | 'pat-warn'; token: string }> {
+  const lines = content.split('\n')
+  const issues: Array<{ file: string; job: string; step: string; kind: 'dead' | 'pat-warn'; token: string }> = []
+
+  // --- 1. Determine if this workflow is push-triggered ---
+  // We look at the `on:` section at root level.
+  let isPushTriggered = false
+  let isBotTriggered = false // workflow_run, workflow_dispatch used by bots
+  let inOnBlock = false
+  let onIndent = -1
```
MickaelV0, line 517: **[nit · nitpick] onIndent variable declared and assigned but never read**

`let onIndent = -1` is declared at line 517 and assigned `onIndent = 0` at line 525, but it is never read anywhere else in the function. Appears to be a leftover from an earlier design that tracked indentation depth. Remove it to keep the linter clean.

<sub>↳ verify: `onIndent` is declared (`let onIndent = -1`, line 517) and assigned (`onIndent = 0`, line 525) inside `detectUnsafeTokenInTriggeredWorkflow`, but is never read. The `inOnBlock` boolean plus a `/^\S/.test(line)` regex handles block-exit detection instead; `onIndent` is a dead leftover from an earlier indentation-dept…</sub>

plugins/dev-core/skills/checkup/doctor-github.ts
```diff
@@ -486,6 +486,332 @@ export function detectMissingContentsRead(
...
+const PUSH_GATE_MERGE_THRESHOLD = 5
+const GRACE_DAYS = 14
+
+function repoAgeOk(owner: string, repo: string): boolean {
+  const result = spawnSync(['gh', 'repo', 'view', `${owner}/${repo}`, '--json', 'createdAt', '--jq', '.createdAt'])
+  if (!result.ok || !result.stdout.trim()) return true // assume ok if can't check
+  const created = new Date(result.stdout.trim())
+  const ageMs = Date.now() - created.getTime()
+  const ageDays = ageMs / (1000 * 60 * 60 * 24)
+  return ageDays >= GRACE_DAYS
```
MickaelV0: **[minor · issue] push-gate: sparse-merge repos produce false negatives due to narrow per_page window**

`countMergeCommits` fetches `per_page = n*3 = 15` commits and counts those starting with "Merge". On repos where merges are sparse relative to total commits (e.g., 5 merges across 50 commits), the function returns < 5 even though >=5 merges have landed, so `mergeCount < PUSH_GATE_MERGE_THRESHOLD` is true and the dead gate is silently skipped. This is a false negative: real dead gates are missed on repos with bursty non-merge commit activity. Fix: increase window to `n * 10` (50 for threshold=5) or use `per_page=100` and accept the API cost.

<sub>↳ verify: Confirmed in the branch code at doctor-github.ts:697. `countMergeCommits` calls the GitHub API with `per_page = n * 3 = 15` (for `n = PUSH_GATE_MERGE_THRESHOLD = 5`). The `--jq` filter counts only those 15 commits that start with "Merge", then the caller at line 794 skips flagging a dead gate if that count is < 5. O…</sub>

plugins/dev-core/skills/checkup/doctor-github.ts
```diff
@@ -486,6 +486,332 @@ export function detectMissingContentsRead(
...
+    for (let s = 0; s < stepHeaders.length; s++) {
+      const stepStart = stepHeaders[s].start
+      const stepEnd = s + 1 < stepHeaders.length ? stepHeaders[s + 1].start : jobLines.length
+      const stepBody = jobLines.slice(stepStart, stepEnd).join('\n')
+
+      // Skip if it uses a safe App token — this step is the mint step itself or consumes app output
+      if (SAFE_PATTERN.test(stepBody)) continue
+
+      for (const pat of DEAD_PATTERNS) {
+        if (pat.test(stepBody)) {
```
MickaelV0, line 637: **[minor · issue] SC3 grace window guards repo age, not workflow age: new workflow on old repo can be falsely flagged**

`repoAgeOk` checks whether the repo is older than `GRACE_DAYS=14`. A workflow file added yesterday to a 3-year-old repo passes the grace check and is evaluated against the full merge history, potentially triggering a false-positive dead-gate flag because the workflow has never had a chance to accumulate push runs. No test covers this case (the PR explicitly discloses SC3 has no unit test). Fix: after `pushCount === 0` is confirmed, also check the workflow's own `created_at` via `gh api repos/{owner}/{repo}/actions/workflows/{wf}` and skip if the workflow is younger than `GRACE_DAYS`.

<sub>↳ verify: Confirmed from branch code. `repoAgeOk` (line 651-658) checks only the repository's `createdAt` date against GRACE_DAYS=14. The SC3 loop at lines 772-800 returns early only when the whole repo is younger than 14 days; it has no per-workflow age guard. `STANDARD_WORKFLOWS` = ['ci.yml', 'auto-merge.yml', 'pr-title.yml…</sub>

plugins/dev-core/skills/checkup/doctor-github.ts
```diff
@@ -486,6 +486,332 @@ export function detectMissingContentsRead(
...
+          const tokenMatch = stepBody.match(pat)
```
MickaelV0, line 638: **[nit · nitpick] repoAgeOk fail-open direction is correct but comment is misleading**

`repoAgeOk` returns `true` (proceed to check) when the API call fails (`!result.ok || !result.stdout.trim()`). The inline comment says `// assume ok if can't check` which is ambiguous: "ok" could mean "grace window ok" (skip) or "age check ok" (proceed). The behaviour is to PROCEED, not to suppress. Suggest: `// can't determine age, proceed with check`.

<sub>↳ verify: Confirmed in the branch at line 656: `if (!result.ok || !result.stdout.trim()) return true // assume ok if can't check`. The call site at line 772 does `if (!repoAgeOk(...))` — so returning `true` means PROCEED. The comment "assume ok if can't check" is genuinely ambiguous: "ok" could be read as "grace-window ok" (i…</sub>

plugins/dev-core/skills/checkup/doctor-github.ts
```diff
@@ -486,6 +486,332 @@ export function detectMissingContentsRead(
...
+            status: 'fail',
+            detail: `push/bot-triggered step uses ${issue.token} — DEAD GATE (GitHub drops pushes from GITHUB_TOKEN; workflow never re-triggers). Use App token or secrets.PAT.`,
+          })
+        } else {
+          checks.push({
+            name: `${fileName}:${issue.job}:${issue.step}`,
+            status: 'warn',
+            detail: `step uses secrets.PAT — legacy token (retiring org-wide). Migrate to App token (actions/create-github-app-token).`,
+          })
+        }
```
MickaelV0, line 810: **[minor · nitpick] push-gate pass entry silently absent when token-taxonomy checks already populate the array**

The `if (checks.length === 0)` guard adds `push-gate history: pass` only when no checks exist yet. If Section A (token taxonomy) already added entries, Section B's loop may complete with zero new entries and no pass row is emitted for push-gate health. JSON consumers iterating check-by-check would see token taxonomy results but silently no push-gate result. Consider tracking A/B check counts separately or unconditionally pushing a push-gate pass when B's loop finds nothing.

<sub>↳ verify: Confirmed real. At line 803 (`if (checks.length === 0)`), the guard tests the shared `checks` array populated by both Section A and Section B. When Section A (token taxonomy) adds any entry — even just the nominal `pass` row at line 702 — `checks.length` is already ≥1 when Section B's loop exits with zero new entrie…</sub>

plugins/dev-core/skills/checkup/__tests__/dead-gate.test.ts
```diff
@@ -0,0 +1,410 @@
...
+  afterEach(() => {
+    fs.rmSync(tmpDir, { recursive: true, force: true })
+  })
+
+  it('AC-5: Dead Gates section appears in output', () => {
+    const result = runDoctor(tmpDir)
+    expect(result.stdout).toContain('Dead Gates')
+  })
+
+  it('AC-4 + AC-5: flags push-gate dead gate in JSON output', () => {
```
MickaelV0: **[major · issue] AC-4 push-gate history path is not isolated — hasFail proves nothing**

The test writes `ci.yml` with `secrets.GITHUB_TOKEN` (fires AC-1 token taxonomy) AND the shared `beforeEach` injects 0 push-runs / 7 merges (fires AC-4 push-gate history). The assertion is `hasFail = true`. This passes if either path trips a fail — the push-gate history branch is never isolated. AC-4 is unproven: a regression that silently disables `countPushRuns` / `countMergeCommits` would leave the test green because the token taxonomy fail still satisfies `hasFail`. Fix: either (a) use a workflow with a safe token (app token) so only the push-gate history path can fire, or (b) assert specifically on a check whose `detail` contains the push-gate sentinel string (e.g., `workflow is a DEAD GATE`) to distinguish it from the token taxonomy sentinel.

<sub>↳ verify: Confirmed from the branch source. The test `'AC-4 + AC-5: flags push-gate dead gate in JSON output'` (dead-gate.test.ts:325) writes a workflow with `secrets.GITHUB_TOKEN` (triggers AC-1 token taxonomy → `status:'fail'`) AND relies on the `beforeEach` 0-push-runs/7-merges setup (intended to trigger AC-4 push-gate his…</sub>

plugins/dev-core/skills/checkup/__tests__/dead-gate.test.ts
```diff
@@ -0,0 +1,410 @@
...
+    const parsed = JSON.parse(result.stdout) as Array<{
+      name: string
+      checks: Array<{ status: string; detail?: string }>
+    }>
+
+    const deadGatesSection = parsed.find((s) => s.name === 'Dead Gates')
+    expect(deadGatesSection).toBeDefined()
+
+    // Token taxonomy check must NOT contain a 'dead' fail about this workflow
+    const tokenFails = deadGatesSection?.checks.filter((c) => c.status === 'fail' && c.detail?.includes('DEAD GATE'))
```
MickaelV0: **[minor · issue] AC-3 no-false-positive filter matches both fail types — filter is not exclusive**

The filter `c.detail?.includes('DEAD GATE')` is used to assert no token-taxonomy fail exists for the App-token workflow. However, the push-gate history fail detail also contains `'DEAD GATE'` (`"workflow is a DEAD GATE (never actually runs on pushes to…)"`). The shared `beforeEach` configures 0 push-runs for all queries; `auto-merge.yml` is absent from the temp repo so the standard-workflows loop may not fire, but the filter's non-exclusivity means a push-gate false-positive would not be caught. Replace with a sentinel unique to the token taxonomy path, e.g., `c.detail?.includes('GitHub drops pushes from GITHUB_TOKEN')`, so the assertion proves specifically that the token classifier did not misfire.

<sub>↳ verify: Confirmed genuine. `STANDARD_WORKFLOWS` (from `github-infra.ts`) includes `'auto-merge.yml'`. The `beforeEach` fake `gh` returns 0 push runs for all `run list --event push` queries and 7 merge commits — both thresholds pass — so the push-gate history check fires for `auto-merge.yml` on staging AND main, emitting fai…</sub>

Issue #3, opened 2026-03-01
feat(cv): add multilanguage support (FR/EN)


MickaelV0:
## Problem

The CV plugin currently supports a single language per invocation. The 2ndBrain project has a working bilingual (FR/EN) fork with:
- Bilingual `cv_data.json` schema (each text field has `fr`/`en` keys)
- `|t` Jinja2 filter for translation switching
- `--lang all` flag to generate all languages at once
- Output: `CV_Name_ATS_FR.md`, `CV_Name_ATS_EN.md`, `CV_Name_Formatted_FR.html`, `CV_Name_Formatted_EN.html`

## Proposed Solution

Add multilanguage support to the upstream cv plugin:

### Data schema
```json
{
  "personal": {
    "name": "John Doe",
    "title": { "fr": "Développeur Senior", "en": "Senior Developer" },
    "summary": { "fr": "...", "en": "..." }
  },
  "experience": [{
    "company": "Acme",
    "role": { "fr": "Lead Dev", "en": "Lead Dev" },
    "highlights": {
      "fr": ["..."],
      "en": ["..."]
    }
  }]
}
```

Fields that are language-neutral (name, company, dates) stay as strings. Translatable fields accept either a string (single-lang) or `{"fr": "...", "en": "..."}` dict (multi-lang).

### Config (`~/.roxabi-vault/config/cv.json`)
```json
{
  "default_language": "fr",
  "supported_languages": ["fr", "en"],
  "default_format": "md"
}
```

### CLI
- `--lang fr|en|all` (default: from config)
- `--format md|html|all`

### Templates
- Jinja2 translate filter: `{{ field|t(lang) }}` — returns `field[lang]` if dict, `field` if string
- One template per format (md, html), language handled by filter

## Context

This is part of the 2ndBrain → roxabi-plugins migration. The local 2ndBrain fork has a working implementation that can serve as reference.

## Acceptance Criteria

- [ ] `cv_data.json` supports both single-lang strings and multi-lang dicts
- [ ] `generate_cv.py` accepts `--lang` flag
- [ ] Templates use `|t` filter for translatable fields
- [ ] Backward compatible: existing single-lang data works without changes
- [ ] `vault-init` or `cv-init` creates example with multi-lang fields

PR #297, opened 2026-06-12
feat(checkup): add per-job CI dormancy detector (#288)

### LINKED ISSUE #288: checkup: conditional-job skip-rate detector (dead-gate per-job dormancy, #579 class)

MickaelV0:
## Summary
- Add sub-check **C** (per-job dormancy) to `checkDeadGates()` in `doctor-github.ts`: cross-references static job wiring (`parseWorkflowJobs`) against per-job run history (`fetchJobHistory` → `aggregateJobStats`) to flag CI jobs that were eligible to run (their workflow ran) yet executed **0 times** across the observation window — escalating to `fail` when the dormant job is also a **required status check** (the #579 signature: a required e2e job filtered out every run while regressions merged under a green check). Renders sub-grouped by job inside the existing `Dead Gates` section; read-only.
- Verdict ladder (`classifyDormancy`, pure): `considered < MIN` → alive · `executed > 0` → alive · else required → `fail` (dormant_required) · empty-matrix/no-`if:` → `warn` (dormant_wiring) · conditional `if:` + not-required → unflagged. Required contexts resolved (`parseRequiredContexts`) by unioning classic `contexts[]` + `checks[].context` + ruleset rules, fail-open on 404.
- **Bug fix surfaced by a de-vacuumed test:** `gh api` rejects `--repo` (`unknown flag`); `fetchRequiredContexts` wrongly passed it → both calls failed-open → required contexts always empty → every dormant *required* job mis-graded `warn` instead of `fail`, **silently disabling the #579 escalation in production**. `--repo` is retained only on the three `gh run list`/`gh run view` calls (where it is valid). A source-inspection test guards against regression.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #288: checkup conditional-job skip-rate detector (#579 class) | OPEN (size:F-lite, P1-high) |
| Analysis | — | Absent (F-lite — analyze skipped) |
| Spec | [288-…-spec.mdx](artifacts/specs/288-conditional-job-skip-rate-spec.mdx) | Present |
| Implementation | 1 feat + 3 docs commits on `feat/288-conditional-job-skip-rate` | Complete |
| Verification | Lint ✅ · Typecheck ✅ · Tests ✅ (dead-gate 18→47, +29) | Passed |

## Test Plan
- [ ] `/checkup` on a repo whose required job skips every run → `fail` under Dead Gates, sub-grouped `wf:branch:job dormancy`
- [ ] Conditional rare-path job (`if:`, not required, skips every run) → **not** flagged (no false positive)
- [ ] Unconditional / empty-matrix job skipping every run → `warn` (dead wiring)
- [ ] Grace window: workflow failing `workflowAgeOk` or job with `< DORMANCY_MIN_RUNS` history → not flagged
- [ ] Required-context lookup works against live `gh api` (no `--repo`) — `dormant_required` escalation fires

Fixes #288

---

### DIFF
--- a/artifacts/frames/288-conditional-job-skip-rate-frame.mdx
+++ b/artifacts/frames/288-conditional-job-skip-rate-frame.mdx
@@ -1,0 +1,86 @@
+---
+title: "checkup: conditional-job skip-rate detector (dead-gate per-job dormancy, #579 class)"
+issue: 288
+status: approved
+tier: F-lite
+date: 2026-06-12
+---
+
+## Problem
+
+`dev-core:checkup` gained a dead-gate detector in #278 (`checkDeadGates()` in `doctor-github.ts`):
+it confirms a workflow file *runs on push* and its token is safe. But detection stops at the
+**workflow** level. A gate can be alive as a file yet contain a **job** that never executes — a
+`paths:` filter that never matches, an `if:` that is permanently false, an empty matrix leg, a
+silently-skipped required check. #278 sees the workflow run, reports the gate healthy, and the
+dormant job is invisible.
+
+This is the exact signature of #579: the e2e suite existed, its workflow ran, the e2e **job** was
+filtered out every time, and 4 regressions shipped under a green check. Workflow-level liveness is
+necessary but not sufficient; a required status check that never actually runs is a merge gate that
+proves nothing.
+
+This frame adds a **per-job dormancy pass** (`detectDormantJobs()`) that cross-references static job
+wiring against per-job run history to flag jobs that were eligible to run but executed 0 times across
+the observation window — escalating when the dormant job is also a required status check.
+
+## Who
+
+- **Primary:** Roxabi maintainers running `/checkup` (manual + cron) on the plugins repo — they get a
+  new finding class under the existing `Dead Gates` section, sub-grouped by job.
+- **Secondary:** any downstream consumer of `dev-core:checkup` on a repo with conditional CI jobs
+  (path filters, matrix, `if:` guards) — the most common place a real #579 hides.
+
+## Constraints
+
+- **Read-only** — static YAML parse + `gh run` history queries only; no execution, no mutation. Cron-safe.
+- **Reuse #278, do not duplicate** — `STANDARD_WORKFLOWS` iteration + the exported grace-window helpers
+  (`repoAgeOk` / `workflowAgeOk`) + the merge-relative anchoring already live in `doctor-github.ts`.
+  `detectDormantJobs()` lands *after* `checkDeadGates()` + `detectUnsafeTokenInTriggeredWorkflow()`.
+- **Eligibility-aware** — a job whose `paths:` simply weren't touched in-window is NOT dormant; only a
+  job that was eligible (its workflow ran, its filter could have matched) yet never executed counts.
+- **Grace window** — jobs younger than the window are skipped (same anchoring as #278); no flag on a
+  brand-new or genuinely-rare-path job.
+- **Surface** — renders in the same `Dead Gates` section, sub-grouped by job; no new top-level section.
+- **F-lite footprint** — single file extended (`doctor-github.ts`) + test cases in the existing
+  `__tests__/dead-gate.test.ts`. No new module, no schema change.
+
+## Out of Scope
+
+- **Workflow-level dormancy** — owned by #278 (`checkDeadGates`); this pass assumes the workflow runs.
+- **Token / push-gate sub-checks** — already shipped in #278.
+- **Executing or "fixing" jobs** — detection only; remediation is the maintainer's call.
+- **Non-standard workflows** — limited to `STANDARD_WORKFLOWS` iteration #278 already established.
+- **Deep `if:` expression evaluation** — we infer dormancy from *observed run history*, not by
+  statically proving an `if:` is always false (history is the ground truth; static parse only
+  enumerates the jobs + their eligibility inputs).
+
+## Premise Validity
+
+**Success in 6 months:** `/checkup` on a repo with a conditional CI job that never runs (path filter
+that never matches, permanently-false `if:`, empty matrix leg) flags that job as dormant under
+`Dead Gates`, sub-grouped by job — and escalates it when the job is also a required status check.
+At least one real dormant-job instance is surfaced on the plugins repo's own CI (or a deliberate
+fixture proves the path end-to-end).
+
+**Failure in 6 months:** a second #579 ships — a required-check job stays dormant and a regression
+merges under a green check while `/checkup` runs and reports nothing — OR the detector is so noisy
+(false-flagging genuinely-rare-path jobs) that maintainers mute the `Dead Gates` section, which is
+boolean-observable: section muted/ignored, or a #579-class incident recurs, within 6 months.
+
+**Simplest alternative:** ship nothing further; rely on #278's workflow-level liveness check.
+**Why not simplest:** #278 is structurally blind to job-level dormancy — it confirms the *file* ran,
+not that each *job* ran. The #579 incident occurred with a workflow that ran fine; only the job was
+dead. The simplest version cannot see the exact failure mode that motivated the parent issue.
+
+## Complexity
+
+**Tier: F-lite** — clear scope, single domain (checkup CI doctor), extends one existing function in
+one file plus its test file; no new architecture, no cross-plugin surface.
+
+Signals observed:
+- `size:F-lite` label on the issue.
+- Single file extended (`doctor-github.ts`) + existing test file (`dead-gate.test.ts`).
+- Reuses established #278 scaffolding (`STANDARD_WORKFLOWS`, grace-window helpers) — no new patterns.
+- One genuine unknown (per-job run-history shape from `gh run view --json jobs` + eligibility
+  modeling) → spec-worthy, but bounded; not multi-domain.
--- a/artifacts/plans/288-conditional-job-skip-rate-plan.mdx
+++ b/artifacts/plans/288-conditional-job-skip-rate-plan.mdx
@@ -1,0 +1,237 @@
+---
+title: "Plan: conditional-job skip-rate detector (per-job dormancy)"
+issue: 288
+spec: artifacts/specs/288-conditional-job-skip-rate-spec.mdx
+complexity: 5/10
+tier: F-lite
+generated: 2026-06-12
+---
+
+## Summary
+
+Extend `checkDeadGates()` in `plugins/dev-core/skills/checkup/doctor-github.ts` with sub-check **C**
+(per-job dormancy): a pure static job parser + a pure verdict function + pure history/required-context
+aggregators wrapped by thin `gh` fetchers, orchestrated by `detectDormantJobs()` and rendered into the
+existing `Dead Gates` section. Single source file + co-located tests in `__tests__/dead-gate.test.ts`.
+
+## Architecture
+
+### Data flow
+
+```mermaid
+flowchart TD
+  subgraph yaml[".github/workflows/*.yml"]
+    WF["workflow YAML (read by checkDeadGates)"]
+  end
+  subgraph src["doctor-github.ts"]
+    N1["parseWorkflowJobs(content) → ParsedJob[]"]
+    AGG["aggregateJobStats(runs, jobs) → Map<job,JobRunStats> (pure)"]
+    N2["fetchJobHistory(owner,repo,wf,branch,jobs,limit)"]
+    PRC["parseRequiredContexts(classic,ruleset) → Set (pure)"]
+    N3["fetchRequiredContexts(owner,repo,branch)"]
+    N6["classifyDormancy(job,stats,isRequired,minRuns) → DormancyVerdict (pure)"]
+    N4["detectDormantJobs(ghOk,owner,repo) → Check[]"]
+    N5["checkDeadGates() — Section C append"]
+  end
+  GH["gh run list / gh run view --json jobs,conclusion --repo o/r"]
+  PROT["gh api required_status_checks + rules/branches (--repo o/r)"]
+
+  WF --> N1 --> N4
+  GH --> N2 --> AGG --> N4
+  PROT --> N3 --> PRC --> N4
+  N4 -->|per job| N6 --> N4
+  N4 -->|"Check[] (fail/warn only)"| N5
+  N5 --> OUT["Dead Gates section → formatText / --json"]
+```
+
+### File × Function map
+
+```mermaid
+flowchart LR
+  subgraph DG["doctor-github.ts"]
+    P1[parseWorkflowJobs]
+    P2[aggregateJobStats]
+    P3[parseRequiredContexts]
+    F1[fetchJobHistory]
+    F2[fetchRequiredContexts]
+    V1[classifyDormancy]
+    O1[detectDormantJobs]
+    C1[checkDeadGates]
+    F1 --> P2
+    F2 --> P3
+    O1 --> P1
+    O1 --> F1
+    O1 --> F2
+    O1 --> V1
+    C1 --> O1
+  end
+  subgraph T["__tests__/dead-gate.test.ts"]
+    UT["unit: parseWorkflowJobs · classifyDormancy · aggregateJobStats · parseRequiredContexts"]
+    IT["integration (subprocess): Dead Gates per-job dormancy"]
+  end
+  P1 -.-> UT
+  V1 -.-> UT
+  P2 -.-> UT
+  P3 -.-> UT
+  C1 -.-> IT
+```
+
+## Agents
+
+| Agent instance | Tasks | Files | Subjects |
+|---|---|---|---|
+| backend-dev-A | T1, T2, T3, T4, T5 | `doctor-github.ts` (+ co-located unit tests in `dead-gate.test.ts`) | parser, verdict, history, required-checks, orchestrator |
+| tester-A | T6, T7 | `dead-gate.test.ts` | integration, coverage |
+
+## Wave Structure
+
+1 wave (sequential chain — all source in one file; parallel split would conflict). ~7 chained tasks.
+Elapsed ≈ sequential (no parallelism gain possible on a single file); 2 named instances for impl/test split.
+
+| Wave | Trigger | Agents | Tasks |
+|------|---------|--------|-------|
+| 1 | start | 1 (backend-dev-A) | T1→T2→T3→T4→T5 |
+| 2 | Wave 1 done | 1 (tester-A) | T6→T7 |
+
+### Budget — per task
+
+| Task | Items | Class | Est. ops | Split? |
+|------|-------|-------|----------|--------|
+| T1 parseWorkflowJobs + tests | 1 | judgmental | 5 | — |
+| T2 classifyDormancy + tests | 1 | judgmental | 5 | — |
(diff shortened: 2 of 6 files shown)


MickaelV0:
## Code Review — PR #297 (per-job CI dormancy detector)

6 fresh domain reviewers (security · backend · tester · devops · architect · product-lead). Single chunk. Verdict: **Request changes** — a small cluster of cheap correctness fixes on the core AC-7 escalation path, plus two test-robustness hardenings.

### 🔴 Should-fix (correctness on the escalation path)

**issue(blocking): `isRequired` ignores matrix-leg required contexts — AC-7 gap for matrix jobs**
`doctor-github.ts:1146`
Root cause: `requiredContexts.has(job.displayName)` matches only the bare static name. A matrix job registers per-leg check contexts (`build (ubuntu-latest)`), never the base `build` — so a dormant *required matrix* job grades `dormant_wiring` (warn) instead of `dormant_required` (fail). The comment at :1145 and spec §Expected-Behavior step 3 both promise leg matching; the code doesn't do it. (The literal #579 single-job case IS handled; this is the adjacent matrix case.)
Fix: `requiredContexts.has(job.displayName) || [...requiredContexts].some(c => c.startsWith(job.displayName + ' ('))`
Confidence: 90% (architect + backend agree; product-lead's "legs aren't required contexts" is incorrect for matrix jobs).

**issue(correctness): in-progress `conclusion: null` counted as executed → fail-open miss**
`doctor-github.ts:931` (source at `:1080`)
Root cause: `j.conclusion !== 'skipped'` is `true` for `null` (GitHub returns `null` for in-progress jobs; unlike `runConclusion` at :1078, the job mapping at :1080 doesn't coerce). A required job skipped 4× + in-progress on the latest run → counted executed → `alive` → the dormant gate is missed.
Fix: `j.conclusion && j.conclusion !== 'skipped'` (or coerce `?? ''` at :1080).
Confidence: 85%.

**issue(correctness): matrix-leg prefix can collide with a parenthesised static job name**
`doctor-github.ts:926`
Root cause: `startsWith(displayName + ' (')` also matches a *static* job literally named `Build (debug)` under a sibling `Build` job → stats cross-contaminate.
Fix: anchor the closing paren — `j.name === displayName || (j.name.startsWith(displayName + ' (') && j.name.endsWith(')'))`; add a regression test (a static `Build (debug)` job is not bucketed under `Build`).
Confidence: 75% (lower-frequency edge).

### 🟡 Test robustness (new tests)

**issue: `bodyOf` source-inspection helper misses `export async function`**
`dead-gate.test.ts:1162`
If `fetchJobHistory`/`fetchRequiredContexts` ever become `async`, `indexOf('export function …')` returns −1 (or the terminator regex over-reads) → the `--repo` assertions pass vacuously. Add `async ` to both the search and the terminator regex.

**suggestion(blocking): base fake-gh glob `*databaseId*` conflates the two `run list` shapes**
`dead-gate.test.ts:384,470`
The `beforeEach` fake answers both `--json databaseId --jq length` (scalar) and `--json databaseId,conclusion` (array) with `echo "0"`. Benign today (all base-fake tests assert 0 push runs), but a structural trap for the next author. Discriminate on `--jq`/`length` like the AC-10 fake does.

### 🔵 Notes (non-blocking)
- `repoAgeOk` runs twice per `checkDeadGates` (`:1241` guard + `:1102` inside `detectDormantJobs`) — one redundant `gh repo view`. Keep the inner guard for standalone safety, or drop it (caller already gated). (architect)
- `wfPath` is cwd-relative (`:1122`) — consistent with the existing doctor design (runs from repo root); fail-open if cwd differs. (devops)
- `checks.length === 0` fallback is effectively unreachable (sub-check A always emits) — pre-existing #278 structure, harmless. (architect)
- `dormant_wiring` detail conflates "broken needs" vs "empty matrix" for the `!hasIf` branch — cosmetic. (devops)
- Spec clean-state wording ("single pass covering B+C") vs impl reality — spec-doc nit, not a behaviour defect. (product-lead)
- `doctor-github.ts` ~1340 lines — natural future split into `…-dormancy.ts`. (architect)
- gh call budget worst-case ~108/run, bounded + short-circuited — fine for cron. (devops)

### 🟢 Praise
- Fail-open is airtight across every `gh`/parse path — a missing/404/malformed protection response never escalates a job to `fail`. (security, devops, product-lead)
- `--repo` discipline correct: present on all `gh run list/view`, absent on `gh api` (the fixed bug) — with a source-inspection regression guard. (security, devops)
- All `spawnSync` calls use argument-array form — no shell-injection surface; parser regexes are ReDoS-safe. (security)
- AC-7 verdict precedence (`isRequired` before `hasIf`) correctly implements the #579 "required + skipped-every-run" case. (product-lead)
- `aggregateJobStats(job, runs)` per-job decomposition is cleaner than the spec's `(runs, jobs) → Map` shape. (architect)

**Verdict: Request changes** — 5 fixes (3 correctness + 2 test). All surgical; none change the public shape.

MickaelV0:
## Review Fixes Applied

All 5 accepted findings from the [code review](#issuecomment) applied (pre-approved as a batch at the review's Phase 8 "Fix now" decision). **Applied:** 5 · **Deferred:** 0 · **Skipped:** 0 · **Failed:** 0.

Commits: `b6fba4f` (findings) + `8b3ba35` (test-flake hardening).

### 🔴 Correctness
- **[applied] #1 — `isRequired` matrix-leg gap** (`doctor-github.ts`): a matrix job's required status checks register one **leg** context each (`Build (ubuntu-latest)`), never the bare parent name — so dormant required matrix jobs graded `warn` instead of `fail`, defeating the AC-7 (#579) escalation. Extracted the inline check into an exported pure **`isJobRequired()`** that also matches any anchored-prefix leg context. Unit-tested directly (4 cases incl. no-false-positive on `Build-extras`).
- **[applied] #2 — null/empty conclusion counted as executed** (`aggregateJobStats`): an in-progress run reports a `null` conclusion; `!== 'skipped'` was truthy for it → fail-open miss of a dormant gate. Now requires a real, non-skipped conclusion (`Boolean(j.conclusion) && …`). Regression: `AC-8f`.
- **[applied] #3 — matrix-prefix sibling collision** (`aggregateJobStats`): a separate static job literally named `Build (debug)` was absorbed into matrix parent `Build`, masking its dormancy. `aggregateJobStats` now takes the set of sibling display names and excludes exact matches from prefix attribution; `fetchJobHistory` threads them in (default `∅` → backward-compatible). Regression: `AC-8g` (sharply demonstrates the masking).

### 🟡 Test robustness
- **[applied] #4 — `bodyOf` helper** now matches `export async function` (and stops the body at the next async export) so the `--repo` source-inspection guard can't silently mis-scope.
- **[applied] #5 — fake-gh `run list` globs** anchored on `--jq` (scalar count) vs `databaseId,conclusion` (history array) so the two distinct shapes can never collide in a future fixture; added a defensive array-shape handler.

### 🔧 Flake hardening (bonus)
- **[applied] AC-1 read-only test** rewritten from a two-stdout byte-comparison to a direct **non-mutation snapshot** (fixture tree before/after a single run). The old assertion flapped under the 4-way parallel pre-push because doctor's live env probes (`which trufflehog`, `command -v roxabi`) flip status under the concurrent trufflehog scan's CPU/IO load — without any state mutation. Now deterministic by construction.

### Verification
`bun run lint` ✅ · `tsc --noEmit` ✅ · vitest **552 passed / 4 skipped** (+7 new tests). AC-1 confirmed stable across repeated runs under concurrent load.

— `/fix` (dev-core)

MickaelV0:
## Code Review — iteration 2 (re-review after fix)

Fresh-agent panel (backend-dev · tester · architect · security-auditor) on the fix delta (`92e83f2…HEAD`). The 5 iteration-1 findings are confirmed fixed; verified the fixes **compose correctly** (no interaction bug between the three `aggregateJobStats` changes) and the new regression tests are **non-vacuous** (each fails if its fix is reverted). Three new findings — all minor; two are cheap correctness/hygiene wins.

### 🔴 Should-fix
1. `issue: isJobRequired sibling asymmetry` — `doctor-github.ts:1113`. `isJobRequired("Build", {"Build (debug)"})` returns `true` because `"Build (debug)".startsWith("Build (")`. If a **separate** static job literally named `Build (debug)` is a required context while the matrix parent `Build` is **not** required, the dormant parent is mis-graded `dormant_required` → false `fail`. This is the exact mirror of fix #3 (sibling collision) — `aggregateJobStats` gained a `siblings` guard but `isJobRequired` did not, so the two sibling-handling sites are now inconsistent. **Fix:** thread the same sibling set in — `requiredContexts.has(d) || [...requiredContexts].some(c => c.startsWith(d + ' (') && !siblings.has(c))`. (backend-dev 78%)
2. `issue: RunRecord.conclusion type lie` — `doctor-github.ts:906`. Typed `conclusion: string`, but `fetchJobHistory:1089` maps it straight from the API where GitHub returns `null` for in-progress jobs → runtime `null` flows through a `string` type. The new `Boolean(j.conclusion)` guard makes it *work*, but a future consumer reading it as a guaranteed string breaks. **Fix:** widen to `string | null` + coerce `?? ''` at fetch; this also lets `AC-8f` use the real `null` shape instead of `''` (tester's same point). (backend-dev 85% · tester 85%)

### 🟡 Test robustness (non-blocking)
3. `issue: AC-1 snapshot covers only tmpDir, not HOME` — `dead-gate.test.ts:563`. The rewrite verifies no mutation of the repo working dir (correct for "does not mutate repo") but passes the **real** `HOME` to the subprocess, so a hypothetical `~/.roxabi-vault` write would escape the assertion *and* pollute real HOME. Doctor is a read-only diagnostic so likely no HOME writes today, but a synthetic HOME (2nd tmpdir, asserted empty) would make it hermetic. Worth a follow-up; touches the shared `runDoctor` helper. (tester 80%)
4. `suggestion(non-blocking): AC-1 runs doctor once` — a lazy-init-on-2nd-run write would slip past. Minor. (tester 65%)

### 🔵 Notes (non-blocking / out-of-scope)
- `aggregateJobStats` optional `siblings` default `∅` → a future caller that omits it silently regresses the collision fix. Add a JSDoc `@param` hazard note. (architect 72%)
- Per-job `new Set([...allNames].filter(...))` is O(n²) in job count — negligible at <50 jobs; leave as-is. (architect 85%)
- In-progress run counts toward `considered`+`skipped` → a transient false-dormancy is *possible* if an in-flight run sits in the 10-run window with ≥5 considered. **Pre-existing, not introduced by this fix.** (backend-dev)
- god-file ~1374 lines — pre-existing; track separate extraction of the T1–T5 dormancy detector. (architect 90%)
- `isJobRequired` lacks a test for exact-AND-leg both present (contrived — GitHub never registers the bare parent). (tester 70%)
- AC-10 is **not** vacuous re STANDARD_WORKFLOWS: it asserts a `fail` *is* produced, so a skipped workflow would fail the test, not pass it.

### 🟢 Praise
- **Security clean:** `startsWith` (no regex → no ReDoS); both fixes are **fail-closed** in the correct direction (more likely to escalate a truly dormant required gate); fake-gh fixtures carry no real secrets/injection surface.
- The three `aggregateJobStats` changes (siblings · null-guard · default-∅) **compose correctly** — verified across in-progress-leg + sibling-only + all-sibling scenarios.
- `AC-8g` and the `isJobRequired` suite are **non-vacuous and load-bearing** (the `Build-extras` case pins the anchored-paren invariant).
- `isJobRequired` extraction is consistent with the module's exported-pure-fn pattern; correct placement at the T4/T5 boundary.

---
**Verdict: Request changes** — findings #1 (sibling consistency/correctness) + #2 (type honesty) are cheap and worth one more pass; everything else is non-blocking. Common-case behavior is correct and well-tested.

*Note: this is review iteration 2 — per the dev-core loop cap, one fix cycle remains before the next review must merge-or-stop.*

🤖 dev-core `/code-review`

MickaelV0:
## Review Fixes Applied

Iteration 2 — both should-fix findings + the JSDoc hazard note, applied as lead (surgical, ~52 LOC across the two files). Commit `f5bbaa5`.

**Applied (user-accepted at Phase 8):** 2 finding(s) + 1 note

### Applied
- [applied] 🔴 `issue: isJobRequired sibling asymmetry` (`doctor-github.ts`) — threaded `siblings` (other jobs' display names) into `isJobRequired`; exact sibling matches are excluded from the prefix match. Mirror of fix #3 (`aggregateJobStats`). A required static `Build (debug)` no longer makes a dormant matrix `Build` read as required → escalate to `fail`. Call site in `detectDormantJobs` computes + passes the sibling set (same idiom as `fetchJobHistory`).
- [applied] 🔴 `issue: RunRecord.conclusion type lie` (`doctor-github.ts`) — widened `conclusion` to `string | null`, the honest shape from `gh run view --json jobs` for in-progress legs. `Boolean(c)` already treats `null` and `''` identically → zero behavior change, only type honesty. `AC-8f` now exercises the real `null` shape alongside `''` and `'skipped'`.
- [applied] 🔵 `thought: aggregateJobStats optional-param silent-regression` — added explicit `@param siblings` JSDoc on **both** `aggregateJobStats` and `isJobRequired` documenting that the empty default reproduces greedy attribution and the dormancy caller MUST pass the real set (default = unit-tests only).

### New regression test
- `isJobRequired` sibling-asymmetry case: `false` when the context is owned by a sibling, greedy-`true` without the guard, still matches a genuine matrix leg with siblings present.

**Deferred / Skipped:** none — the 🟡 (AC-1 HOME hermeticity, single-run) + remaining 🔵 notes (O(n²) sibling set, in-progress transient, god-file) were assessed non-blocking and left for a follow-up sweep, not this PR.

**QG:** lint 0 · typecheck 0 · vitest **553 passed | 4 skipped**.

> Re-review (iteration 3 — final per the 2-iteration loop cap) follows.

MickaelV0:
## Code Review — iteration 3 (final)

Re-review of commit `f5bbaa5` (iteration-2 fixes). Three fresh agents (backend-dev, tester, security-auditor) scoped to the ~52 LOC delta + the two should-fix findings.

**Verdict: ✅ Approve with comments** — both iteration-2 🔴 should-fix findings verified genuinely resolved (non-vacuous, fail-closed); no regression; security clean. Remaining items are non-blocking nitpicks, each assessed "no fix required this iteration".

### ✅ Verified resolved
- **#1 isJobRequired sibling asymmetry** — guard `!siblings.has(c)` correctly excludes a context owned by a static sibling while still matching genuine matrix legs. Call site computes `allJobNames \ {self}` per-job (no off-by-one). Now **symmetric** with the `aggregateJobStats` guard (backend-dev 90%). Fail-closed confirmed: no real matrix leg is ever excluded → no false-negative dead-gate (security-auditor).
- **#2 RunRecord.conclusion type lie** — `string | null` threaded through the sole consumer; `Boolean(j.conclusion)` keeps `null` on the not-executed side (zero behavior change, type honesty only). Fetch coercion asymmetry (`runConclusion ?? ''` vs leg `conclusion` left `null`) is intentional + correct (backend-dev, security-auditor).
- **Tests non-vacuous** — AC-8f fails if the `Boolean()` guard is removed (`null !== 'skipped'` → wrongly executed); the new isJobRequired test's assertions 1+3 fail if `!siblings.has(c)` is removed or over-aggressive (tester).

### 🔵 Non-blocking (notes — no fix required)
- `nitpick: collision-edge symmetry` (`doctor-github.ts:1139/941`, backend-dev 90%) — if a matrix leg name *equals* a static sibling name, both functions symmetrically deny attribution → a spurious `dormant_wiring` warn is *possible*. **Not a regression** (prior code's failure mode was worse: parent absorbs sibling's runs); pathological YAML construction. Internally consistent.
- `nitpick: AC-8f bundles null+'' in one assertion` (`dead-gate.test.ts:1128`, tester 82%) — a future change that fixes `null` but regresses `''` would be invisible (the `null` leg still suppresses the count). "Acceptable as written"; isolation gap is low-severity. Split into two `it` blocks if revisited.
- `note: isJobRequired assertion 2 documents greedy behavior, not the guard` (`dead-gate.test.ts:1191`, tester 95%) — correct by design (greedy-default doc); a one-line comment would clarify intent.

### 🟢 Praise
- Security clean: fail-closed escalation preserved on both changes; `startsWith` (no regex) → no ReDoS; low-trust `gh`/YAML names reach only comparison + display, no exec/FS/DB sink; no secret leak (security-auditor).
- The two sibling guards are now symmetric; JSDoc on both helpers documents the empty-default silent-regression hazard.

---
**Loop cap:** iteration 3 = final re-review. Remaining nitpicks are non-blocking → recommend merge, with an optional follow-up issue for the two test-isolation nitpicks.

PR #301, opened 2026-06-15
docs: add functional framing rules to decision-presentation Pattern A

MickaelV0:
## What
Enrich the Decision Presentation Protocol (Pattern A) in `plugins/shared/references/decision-presentation.md` with **functional-framing rules**: present the decision in business/user terms (intention, what each option improves/costs, upstream→downstream history), avoid file paths / symbol names / infra jargon unless the user already used the term, and keep implementation detail out of the prompt.

## Why
This preference recurred across sessions and lived only in Claude's auto-memory (`feedback_dp_functional_framing`). Folding it into the shared reference that every DP-emitting skill loads (`→ DP(A)`) makes it structural instead of recall-dependent. The user-preference half is also canon in `bouly-site/brand/PORTRAIT.md` §8.

## Scope
- Single file: `plugins/shared/references/decision-presentation.md` (+9 / −3)
- No skill/command change → no `dev-core` version bump required (per the `check-skill-version` gate, which only covers `plugins/*/skills/**` and `commands/**`).

### DIFF
--- a/plugins/shared/references/decision-presentation.md
+++ b/plugins/shared/references/decision-presentation.md
@@ -19,12 +19,18 @@ Target:      {what we're trying to achieve}
 Path:        {how we get there once a choice is made}
 
 Options:
-  1. {option} — {one-line description}
-  2. {option} — {one-line description}
-  3. {option} — {one-line description}
+  1. {option} — {what it improves} / {what it costs}
+  2. {option} — {what it improves} / {what it costs}
+  3. {option} — {what it improves} / {what it costs}
 
 Recommended: Option N — {1-line rationale}
 ```
+
+**Frame functionally, not technically** — the user is deciding intent and consequences, not reviewing syntax:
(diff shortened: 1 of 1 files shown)


MickaelV0:
## Code Review — light (docs-only)

**Scope:** 1 file, +9/−3 — `plugins/shared/references/decision-presentation.md` (markdown reference, no code paths). Secret scan: clean · behind base: 0.

**Findings**

praise: Consolidates a recurring user-comms preference (functional DP framing) into the canonical `→ DP(A)` reference that every skill loads — structural over recall-dependent. Options now read as improves/costs, matching the framing.

nitpick (non-blocking): The jargon examples (`Quadlet, nkey, ACL, JetStream, KV-store`, "NATS KV", "Fernet keyring") are Roxabi-specific while this is a project-agnostic shared reference (Design Principle #1). They read as illustrative → non-blocking, but a future pass could genericize them ("orchestration / message-bus / key-store internals").
  -- light-review
  Confidence: 60%

**Verdict: ✅ Approve with comments** — no blockers. Fence balanced, glosses correct, notation (`¬`, `→`) consistent. CI green, mergeable CLEAN.

PR #337, opened 2026-07-09
fix(dev-core): robust origin/staging base detection

MickaelV0:
Robustness pass on dev-core base-branch detection + consistent base diffing. Reviewed by a 4-agent `/code-review` and a 2-agent re-review; the re-review drove a descope (below). 6 commits.

## Final design

- **`shared/lib.sh` `detect_base_branch`** — returns the first of `staging`, `main`, `master` that exists as a remote-tracking ref (exact `git show-ref --verify`), else `main`. Replaces `git branch -r | grep -q 'origin/staging'`, an unanchored substring match (`origin/staging-x`, or a remote named `myorigin`, false-positived). Result is always one of the three protected branches, so cleanup's protected-set model stays valid.
- **`pr/gather-state.sh`** — `git fetch --prune origin` before detection, and diffs against `origin/${BASE}` (was bare local `${BASE}`, which rev-parse resolves to a possibly-stale/absent local branch first).
- **`cleanup/analyze-branches.sh`** — detects base after `fetch --prune`; `branch_merged()` now compares against a resolved `BASE_REF` (prefer `origin/<base>`, else local, else skip) so an unmerged branch is never mis-classified `safe_delete` when the base isn't checked out locally. **+ regression test.**
- **DRY** — `test`, `code-review`, `setup-worktree` inlined the detection one-liner (already drifted); now source `lib.sh`.
- **Consistency** — `code-review`/`test`/`pr`/`dev` skills diff against `origin/${BASE}`.

## Review → descope

An `origin/HEAD`-based fallback (added mid-PR for `master`-default repos) introduced a class of issues: `git fetch` never refreshes `refs/remotes/origin/HEAD` (dangling-symref → wrong base), and it decoupled cleanup's `{main,master,staging}` protected set. Rather than generalize all of cleanup to arbitrary defaults, detection was **constrained to the triad candidate-list** — keeping `master` support while leaving cleanup's protection model valid. Two interim-fix defects the re-review found (unvalidated fallback; a regex-metachar branch name that could zero the remote-branch list) are gone with that revert.

## Verification (empirical, scratch repos, `set -e`)

| refs present | base |
|---|---|
| `origin/staging` (+ decoys `origin/staging-x`, `myorigin/staging`) | `staging` |
| decoys only | `main` |
| `master`-only repo | `master` |
| `main` + `master` | `main` |
| no origin / empty repo | `main` |

`analyze-branches` suite (existing + new origin-only-base regression) green · `validate_plugins.py` all-pass · typecheck clean · lefthook pre-push `test` green.

## Versions
`dev-core` plugin 0.8.14 → **0.8.15**. Skills: `pr` 0.4.3, `test` 0.4.1, `code-review` 0.2.2, `setup-worktree` 0.3.1, `cleanup` 0.6.1, `dev` 0.3.1.

---
Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/plugins/dev-core/.claude-plugin/plugin.json
+++ b/plugins/dev-core/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "dev-core",
-  "version": "0.8.14",
+  "version": "0.8.15",
   "description": "Full development workflow — frame, shape, plan, implement, review, ship. 30 skills, 9 agents, safety hooks.",
   "author": {
     "name": "Roxabi"
--- a/plugins/dev-core/skills/cleanup/SKILL.md
+++ b/plugins/dev-core/skills/cleanup/SKILL.md
@@ -2,7 +2,7 @@
 name: cleanup
 argument-hint: [--all | --report-only | --yes | --scope <#N>]
 description: Clean git branches/worktrees/remotes after merge-status verification; sweep stuck pipeline labels and orphan CI runs. Triggers: "cleanup" | "clean branches" | "cleanup worktrees" | "remove stale branches".
-version: 0.6.0
+version: 0.6.1
 allowed-tools: Bash, Read, EnterWorktree, ExitWorktree, ToolSearch
 ---
 
--- a/plugins/dev-core/skills/cleanup/__tests__/analyze-branches.test.sh
+++ b/plugins/dev-core/skills/cleanup/__tests__/analyze-branches.test.sh
@@ -15,7 +15,7 @@ ANALYZE="${SCRIPT_DIR}/../analyze-branches.sh"
 
 TMPDIR_FIXTURE="$(mktemp -d)"
 WT_DIR="${TMPDIR_FIXTURE}/wt-i18n"
-trap 'rm -rf "$TMPDIR_FIXTURE"' EXIT
+trap 'rm -rf "$TMPDIR_FIXTURE" "${ORIGIN_ROOT:-}"' EXIT
 
 cd "$TMPDIR_FIXTURE"
 git init -q
@@ -96,5 +96,40 @@ assert_eq "--scope 19 includes only feat/19-auth (excludes #33, #319)" "feat/19-
 result_hash_scoped="$("$ANALYZE" --json --no-fetch --scope "#19")"
 DEBUG_JSON="$result_hash_scoped"
 assert_eq "--scope #19 strips the leading #" "19" "$(echo "$result_hash_scoped" | jq -r '.scope')"
+
+# Regression: base present only as origin/<base>, not checked out locally (the
+# worktree-per-issue norm). branch_merged must resolve origin/<base> and must NOT
+# mis-classify an unmerged branch as safe_delete just because no LOCAL base branch
+# exists. Before the BASE_REF fix, `git log staging..feat/50-thing` errored (no
+# local staging) -> empty stdout -> merged=true -> safe_delete.
+ORIGIN_ROOT="$(mktemp -d)"
+git init -q --bare "${ORIGIN_ROOT}/origin.git"
+git init -q "${ORIGIN_ROOT}/work"
+cd "${ORIGIN_ROOT}/work"
+git config user.email "test@example.com"
+git config user.name "Test User"
+git remote add origin "${ORIGIN_ROOT}/origin.git"
+echo base > f.txt
+git add f.txt
+git commit -q -m "chore: base"
+git branch -M staging
+git push -q -u origin staging
+echo wip > wip.txt
+git checkout -q -b feat/50-thing
+git add wip.txt
+git commit -q -m "feat: wip (#50)"
+git checkout -q -b scratch        # sit off the feature so it is not the current branch
+git branch -q -D staging          # only origin/staging remains as the base ref
+git fetch -q origin               # ensure the origin/staging tracking ref exists
+
+result_origin="$("$ANALYZE" --json --no-fetch)"
+DEBUG_JSON="$result_origin"
+origin_base="$(echo "$result_origin" | jq -r '.base_branch')"
+feat50_action="$(echo "$result_origin" | jq -r '.local_branches[] | select(.name == "feat/50-thing") | .action')"
+feat50_merged="$(echo "$result_origin" | jq -r '.local_branches[] | select(.name == "feat/50-thing") | .merged')"
+
+assert_eq "origin-only base resolves to staging" "staging" "$origin_base"
+assert_eq "unmerged branch NOT safe_delete when base only on origin/*" "unmerged" "$feat50_action"
+assert_eq "unmerged branch merged=false when base only on origin/*" "false" "$feat50_merged"
 
 echo "PASS: analyze-branches.test.sh"
--- a/plugins/dev-core/skills/cleanup/analyze-branches.sh
+++ b/plugins/dev-core/skills/cleanup/analyze-branches.sh
@@ -59,6 +59,18 @@ fi
 # Detect base AFTER fetch --prune so a freshly-created (or remotely-deleted)
 # origin/staging is reflected in the local remote-tracking refs before detection.
 BASE_BRANCH="$(detect_base_branch)"
+# Ref to compare branches against for merge detection: prefer the remote-tracking
+# base (fresh after fetch --prune, and present even when the base branch is not
+# checked out locally — the norm in the worktree-per-issue flow), else a local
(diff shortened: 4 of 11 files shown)


MickaelV0:
## Code Review — PR #337

**Verdict: 🔴 Request changes** — 4 fresh domain agents (security, architecture, devops, tester). Diff clean of secrets, no injection. But the review surfaced **2 confirmed correctness issues introduced by the `origin/HEAD` fallback** (Problem 2) plus consistency drift from applying the `origin/${BASE}` migration to `pr/` only. Both `issue:` findings were reproduced empirically in scratch repos by two independent agents.

### 🚫 Blockers

**1. `issue:` `origin/HEAD` fallback returns a stale/dangling base — `git fetch --prune` does NOT refresh `refs/remotes/origin/HEAD`**
`shared/lib.sh:20` (interacts with `gather-state.sh:13`, `analyze-branches.sh:56`)
`refs/remotes/origin/HEAD` is a symbolic ref updated only by clone / `git remote set-head`, **never by `git fetch`/`--prune`**. Verified: after a server-side default-branch rename + `fetch --prune`, `git symbolic-ref --short refs/remotes/origin/HEAD` still returns the **old** name with exit 0, even when that ref was just pruned. `detect_base_branch` trusts it verbatim → `gh pr create --base <dangling>` fails, or `git log "origin/${BASE}..HEAD" 2>/dev/null || echo none` silently reports an empty diff/commit-log. Only reachable on the no-`origin/staging` path (Roxabi repos are safe), but the lib ships to arbitrary repos. *(devops 90%, tester 70%)*
→ Fix: validate the resolved default (`git show-ref --verify --quiet refs/remotes/origin/<default>`) before trusting it; else fall back to `main`.

**2. `suggestion(blocking):` `cleanup` protected-branch set decoupled from the now-generalized detector**
`cleanup/analyze-branches.sh:63` (+ exclusion regex `:302`, `cleanup/SKILL.md:298`)
This PR widened the fallback from "always `main`" to "whatever `origin/HEAD` points at" (`develop`, `trunk`, …). But `PROTECTED_JSON='["main","master","staging"]'` and the remote-exclusion regex are still the hardcoded triad. On a `develop`-default repo with no `staging`: detector returns `develop` → not protected → `git log develop..develop` empty → `merged=true` → **the repo's default branch is offered for deletion** in `/cleanup` (human-gated, but actively recommended). *(architect 72%)* `Class: target-axis-trap`
→ Fix: derive `PROTECTED_JSON` to include `$BASE_BRANCH` + add it to the `:302` exclusion.

**3. `suggestion(blocking):` `origin/${BASE}` migration applied to `pr/` only — `code-review` + `test` still diff bare local `${BASE}`**
`code-review/SKILL.md:54,55,62`, `test/SKILL.md:59` `Class: parallel-path-drift`
Detection was migrated to `detect_base_branch` in these skills, but consumption still does `git diff ${BASE}...HEAD` with no `origin/` prefix — the exact stale/absent-local-ref failure the PR fixed for `pr/`. In a `.claude/worktrees/…` flow where local `staging` is never checked out, `git diff staging...HEAD` → `fatal: bad revision` → empty Δ → `/code-review` & `/test` falsely halt "nothing to review". *(architect 80%, devops)*
→ Fix: prefix `origin/${BASE}` in those diffs (matching `pr/` and `scan-state.sh`).

### ⚠️ Warnings (pre-existing, amplified by this change)

**4. `issue:` `analyze-branches.sh` `branch_merged()` bare `${BASE_BRANCH}` + `2>/dev/null` treats ref-resolution failure as `merged=true`** — `:150,169,177`. A dangling base (blocker 1) makes branches look merged → `safe_delete`. Pre-existing; recommend a separate follow-up (distinguish "ref failed" from "0 commits ahead"). *(devops 75%, tester)*

**5. `suggestion(non-blocking):` `pr/SKILL.md:56`** "No commits" guard still `git log ${β}..HEAD` (bare) while Step 3a two headings below now uses `origin/${β}` — they can disagree. *(tester 55%)*

**6. `suggestion(non-blocking):` `dev/SKILL.md:89`** spec block hardcodes `origin/staging..HEAD` while `scan-state.sh:89` uses `origin/${BASE}` — doc drift. *(architect 70%)*

**7. `suggestion(non-blocking):` inline `${CLAUDE_SKILL_DIR}` source is unguarded** — if unset, `BASE=""` propagates silently. Low-prob (var normally present). *(architect 55%)*

**8. `nitpick:` `dev/scan-state.sh:89`** reads `origin/${BASE}` but lacks the `git fetch --prune` its sibling `gather-state.sh` gained (may be intentional — hot path). *(architect 45%)*

### ✅ Praise
- `git show-ref --verify` exact-ref idiom correctly kills the substring false-positive (`origin/staging-x`, `myorigin`). *(security, devops)*
- `local default` split from `default=$(...)` correctly sidesteps SC2155 so `|| true` actually guards; function provably exits 0 on all 5 branches. *(devops 95%)*
- detect-after-`fetch --prune` ordering + explicit prefetch correctly close the "freshly-created origin/staging" race. *(architect)*
- Intra-plugin `source lib.sh` is the right DRY mechanism (copy-sync governance is cross-plugin only). No security/injection paths; no secrets. *(security, architect)*

---
*4 agents · single-chunk · findings ranked by confidence · verify empirically before applying.*

MickaelV0:
## Review findings addressed — `64f14dd`

| Finding | Resolution | Verified |
|---|---|---|
| **B1** `origin/HEAD` dangling after `fetch --prune` | `lib.sh` now validates the resolved default exists (`show-ref --verify refs/remotes/origin/<default>`) before returning it, else `main` | ✅ scratch matrix: staging→staging · fresh master→master · **dangling HEAD→main** · no-HEAD→main |
| **B2** `cleanup` protected set decoupled | `PROTECTED_JSON` derived from `$BASE_BRANCH` (`jq … unique`) + `origin/${BASE_BRANCH}$` added to the `:302` exclusion regex | ✅ `BASE_BRANCH=develop` → `["develop","main","master","staging"]`, `origin/develop` excluded from delete candidates |
| **B3** `code-review`+`test` diff bare local `${BASE}` | prefixed `origin/${BASE}` (diffs + secret-scan) | ✅ grep: 0 bare base-ranges remain across all skills |
| **W5/W6** `pr` guard rail + `dev` implement-spec | `origin/` prefix for consistency | ✅ |

Deferred to a separate follow-up (pre-existing, not introduced here): `analyze-branches.sh` `branch_merged()` treats ref-resolution failure as `merged=true` — needs a `rev-parse --verify` guard to distinguish "ref failed" from "0 commits ahead".

Gates: `validate_plugins.py` all-pass · typecheck clean · lefthook pre-push `test` green.

**Verdict → ✅ blockers resolved.** Ready for `reviewed` label once a re-check is satisfied.

MickaelV0:
## Re-review resolved — `97792f4`

The 2-agent re-review of the fix commit found the `origin/HEAD` fallback was the root of the whole issue class (dangling symref, unvalidated `main` fallback R1, `cleanup` protected-set decoupling B2/R4). **Descoped**: `detect_base_branch` now uses a `staging → main → master` candidate-list (each `show-ref --verify`-validated), so it keeps `master` support but always returns a protected-triad member — mooting B1/B2/R1/R4 and reverting the fragile interpolated grep clause (R2).

| re-review finding | resolution |
|---|---|
| R1 `main` fallback unvalidated | moot — candidate-list validates each ref, only bare last-resort `main` |
| R2 interpolated grep zeroes branch list | reverted the `origin/${BASE_BRANCH}$` clause |
| R3 / F4 `branch_merged()` local base | **deferred** — pre-existing, test-coupled, not worsened |
| R4 run-cancellation triad | moot — base always ∈ triad |
| R5 `code-review:21` header | fixed → `origin/${BASE}` |

Verified: candidate-list matrix (staging/decoys→main/master-only→master/main+master→main/no-origin→main) + existing `analyze-branches` test passes unchanged. Gates green.

This iteration is simpler and lower-risk than the version reviewed (trivial candidate-list + reverts to known-good literals).

MickaelV0:
## Deferred follow-up now folded in — `e1ec2ae`

The `branch_merged()` local-base issue (originally deferred) is fixed in this PR after all: it now resolves a `BASE_REF` (prefer `origin/<base>`, else local, else skip). Reproduced the bug and locked it with a committed regression test — base present only as `origin/staging`, no local `staging`, an unmerged feature branch must classify `unmerged` (it classified `safe_delete` on the old path). Existing suite unchanged; new case green.

No open follow-ups remain. Ready for `reviewed`.

PR #339, opened 2026-07-15
fix(dev-core): write lifecycle artifacts as .md not .mdx

MickaelV0:
## Summary
- Lifecycle artifacts (`frame` / `analyze` / `spec` / `plan` / interview / consensus) are written as **`.md`** instead of **`.mdx`**.
- Locate paths dual-read **`.md` then legacy `.mdx`** so existing repo artifacts still resolve.
- Plan skill templates renamed (`plan-template.md`, `micro-task-example.md`); `dev-core` **0.8.15 → 0.8.16**.
- ADRs and Fumadocs docs under `docs/` stay **`.mdx`**. Historical `artifacts/**/*.mdx` not bulk-renamed.

## Test plan
- [ ] `/frame` (or free-text) writes `artifacts/frames/*-frame.md`
- [ ] `/spec` / `/plan` write `*-spec.md` / `*-plan.md`
- [ ] `/dev #N` / `scan-state.sh` still finds existing `.mdx` artifacts for an old issue
- [ ] `gather-state.sh` on a branch with only legacy `.mdx` still emits `analysis=` / `spec=`
- [ ] ADR skill still creates `docs/architecture/adr/{NNN}-*.mdx`


MickaelV0:
## Code Review — PR #339 (write lifecycle artifacts as .md not .mdx)

**Verdict: Request changes** — one blocking defect. The migration is otherwise well-executed and the completeness sweep is genuinely clean (every surviving `.mdx` — ADRs, Fumadocs docs, historical refs — is intentionally out of scope).

### Blocker
- `issue:` **`clarify/SKILL.md` reads strict `.md` and silently drops all legacy `.mdx` reads.** Lines 63 + 76–79 use `artifacts/{frames,analyses,specs,plans}/{N}-*.md` (no trailing `*`). On-disk state of this very repo at `staging`: **81 `.mdx` artifacts** (24 frames, 8 analyses, 24 specs, 25 plans) vs 2 `.md`. So `/clarify --resume --issue N` resolves **zero** frames/specs/plans for every pre-existing issue and degrades to issue-body-only synthesis — a confident, impoverished recap with no signal that prior state was missed. The PR's own body promises dual-read.
  - Fix: `.md` → `.md*` on the 5 lines (matches the 12 sibling `.md*` sites).

### Root cause
- `Class: parallel-path-drift` — the "locate lifecycle artifact by issue number" concern is open-coded **4 different ways** across sibling skills (`.md*` glob ×12, explicit dual operands ×3, ext-agnostic `Glob`/`grep` ×3, strict `.md` ×1). The break landed in the one callsite that didn't *look* like a locate (prose table rows, not `ls`). Recommend stating the convention once in `references/dev-process.md`: `lifecycle locate = {dir}/{N}-*.md*` (trailing `*` = legacy `.mdx` compat, load-bearing).

### Non-blocking
- `suggestion(non-blocking):` `plugin.json` bump `0.8.15→0.8.16` under-signals a write-behaviour change + a new compat contract; `0.9.0` fits better on a 0.x line (version discipline otherwise correct — not set in marketplace entry).
- `nitpick:` `code-review/SKILL.md:77` spec-locate now emits two operands without `head -1` → can bind two paths where Step 3 assumes one. Append `| head -1`.
- `suggestion(blocking→treated non-blocking):` the 12 `.md*` globs carry no marker explaining the load-bearing trailing `*`; a future tidy-pass could normalize it to `*.md` and silently re-break legacy. Add a one-line note per `Let:` block or the single `dev-process.md` note above.
- `thought:` `ls a.md a.mdx | head -1` ordering — `.md` correctly wins for the same slug (the real re-run case). Legacy can only win across *different* slugs, a pre-existing ambiguity identical under `.md*`. No change required for correctness.

PR #342, opened 2026-07-16
fix(ci): gate dependabot majors on fetch-metadata, not the PR title

MickaelV0:
## Problem

The `Block dependabot semver-major bumps` step in `auto-merge.yml` has been **inert since dependabot grouping was enabled**. It parses `from X to Y` out of the PR *title* — a shape only single-dependency Dependabot PRs have. Both ecosystems are grouped (`patterns: ["*"]`), so every Dependabot PR is titled `bump the <group> group across 1 directory with N updates`, with no versions in the title. The regex never matches → the guard silently no-ops.

The workflow's own header comment ("semver-major bumps are never auto-merged — the vite 8 SSR regression stayed dormant 10 weeks") documents a control that does not exist.

Verified against the two open Dependabot PRs, each containing a real major:

| PR | Title | Old regex |
|----|-------|-----------|
| #306 | `bump the github-actions group across 1 directory with 2 updates` (checkout 6→7) | **passes** (no block) |
| #338 | `bump the npm-dependencies group across 1 directory with 6 updates` (typescript 6→7) | **passes** (no block) |

Both are one `reviewed` label away from auto-merging a major.

## Fix (root cause, two layers)

- **`auto-merge.yml`** — replace the title-regex with [`dependabot/fetch-metadata`](https://github.com/dependabot/fetch-metadata) (SHA-pinned `@25dd0e34…` = v3.1.0), gating on `update-type == 'version-update:semver-major'`. For grouped PRs, fetch-metadata reports the **highest** update-type across the group — so a major hidden inside a group is caught, which a title parse never could be.
- **`dependabot.yml`** — exclude majors from both groups (`update-types: [minor, patch]`) so majors arrive as individually-titled, individually-reviewable PRs instead of riding inside a `chore(deps)` group. Bump `semver-major-days` cooldown 3→7 for soak.

## Effect on the open PRs

Once this lands, Dependabot re-evaluates its config and re-emits #306/#338: the safe minor/patch bumps stay grouped, and each major (checkout 6→7, typescript 6→7) breaks out into its own PR — where the guard now blocks auto-merge and forces manual validation. TS 7 in particular breaks this repo's LSP (`typescript-language-server` needs the `tsserver` binary that TS 7 removed) — see the review on #338.

## Test notes

- YAML of both files validated (`yaml.safe_load`).
- `fetch-metadata` only runs on `pull_request` activity where `github.event.pull_request.user.login == 'dependabot[bot]'`; the `check_suite`/`push` triggers have no PR context and skip the step (no error).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,9 +8,14 @@ updates:
       github-actions:
         patterns:
           - "*"
+        # Majors excluded from the group → emitted as individual PRs so the
+        # auto-merge semver-major guard reviews each one on its own.
+        update-types:
+          - minor
+          - patch
     cooldown:
       default-days: 3
-      semver-major-days: 3
+      semver-major-days: 7
       semver-minor-days: 3
       semver-patch-days: 3
   - package-ecosystem: npm
@@ -21,8 +26,13 @@ updates:
       npm-dependencies:
         patterns:
           - "*"
+        # Majors excluded from the group → emitted as individual PRs so the
+        # auto-merge semver-major guard reviews each one on its own.
+        update-types:
+          - minor
+          - patch
     cooldown:
       default-days: 3
-      semver-major-days: 3
+      semver-major-days: 7
       semver-minor-days: 3
       semver-patch-days: 3
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -44,19 +44,31 @@ jobs:
           app-id: ${{ vars.ROXABI_CI_APP_ID }}
           private-key: ${{ secrets.ROXABI_CI_APP_PRIVATE_KEY }}
 
-      - name: Block dependabot semver-major bumps
+      # Read the real update-type from Dependabot's metadata instead of parsing
+      # the PR title. The old title-regex only matched single-dependency titles
+      # ("bump X from A to B"); grouped PRs are titled "bump the <group> group…"
+      # with no versions, so the guard never fired for a grouped major. For a
+      # grouped PR, fetch-metadata reports the HIGHEST update-type across the
+      # group — so a major hidden inside a group is caught.
+      - name: Fetch dependabot metadata
+        id: dependabot-meta
         if: github.event.pull_request.user.login == 'dependabot[bot]'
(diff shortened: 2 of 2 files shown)


MickaelV0:
## Post-merge correction + live verification

Two updates to the record for this PR.

### ✅ The guard is verified — live, on the exact case the old regex missed

Tested by labelling #306 `reviewed` (its payload is verified-safe, so a guard failure would only have merged a safe bump; #338 was deliberately **not** used):

```
Fetch dependabot metadata        success → outputs.update-type: version-update:semver-major
Block dependabot semver-major    failure  ← guard fired, comment posted
Enable auto-merge (merge commit) skipped  ← never enabled
```

`fetch-metadata` correctly reported `semver-major` on a **grouped** PR. The control now exists for real.

### ❌ Correction: this PR body's claim about re-emission is wrong

> ~"Once this lands, Dependabot re-evaluates its config and re-emits #306/#338: the safe minor/patch bumps stay grouped, and each major breaks out into its own PR."~

Not automatic. `@dependabot recreate` on both PRs **kept the majors grouped** (`checkout` 6→7 in #306, `typescript` 6→7 in #338), despite `update-types: [minor, patch]` being live on `staging`.

The config is **correct per the docs** — group `update-types` takes the bare form (`minor`/`patch`); the prefixed `version-update:semver-*` form belongs to `ignore`, not `groups`. And the docs state *"Any outdated dependencies that do not match a rule are updated in individual pull requests."* So the observed behaviour contradicts documented behaviour.

Most consistent explanation with the evidence: `recreate` re-resolves **versions** for a **fixed dependency set** without re-deriving group membership (it bumped an unrelated dep 2.5.3→2.5.4 while keeping the major). `recreate`'s effect on grouping is **undocumented**, so this is inference, not fact.

**Net effect on safety: none — the property holds.** The guard is the load-bearing control and it is proven; the `update-types` split is defense-in-depth whose failure the guard already catches. It should apply to newly-created group PRs going forward.

**To verify it definitively:** Insights → Dependency graph → Dependabot → "Check for updates" (the only documented way to force a fresh job; no API equivalent). Not done — it needs the UI.

Issue #345, opened 2026-07-16
fix(dev-core): init.ts shim resolves dev-init without the version path segment — /env-setup Phase 2 and /init scaffolds are dead


MickaelV0:
## Symptom

Every documented path that calls the init CLI dies immediately:

```
$ bun <dev-core>/skills/init/init.ts scaffold-rules --stack-path .claude/stack.yml --claude-md CLAUDE.md
dev-init plugin not found — install dev-init from roxabi-marketplace (sibling of dev-core).
```

`dev-init@roxabi-marketplace` **is** installed (`claude plugin list` confirms it). The message is wrong: it is a path-resolution bug, not a missing plugin.

This kills `/env-setup` **Phase 2 (Scaffold CLAUDE.md Critical Rules)** and every `/init` path that shells into `scaffold-rules` / `scaffold-docs` / `scaffold-fumadocs`.

## Root cause

`plugins/dev-core/skills/init/init.ts` is a shim that probes two candidates:

```ts
join(dirname(root), 'dev-init', 'skills', 'init', 'init.ts'),   // line 12
join(root, '..',     'dev-init', 'skills', 'init', 'init.ts'),  // line 13
```

with `root = CLAUDE_PLUGIN_ROOT`. But the plugin cache is **version-scoped**, so `root` already ends in a version segment:

```
~/.claude/plugins/cache/roxabi-marketplace/dev-core/0.8.16          <- root
~/.claude/plugins/cache/roxabi-marketplace/dev-init/900722eace17    <- actual sibling
                                                    ^^^^^^^^^^^^ never accounted for
```

Both candidates resolve to `…/roxabi-marketplace/dev-init/skills/init/init.ts`, which does not exist — the version directory is missing from the path.

## Evidence

```bash
$ ls ~/.claude/plugins/cache/roxabi-marketplace/dev-init/
900722eace17

$ ls ~/.claude/plugins/cache/roxabi-marketplace/dev-init/skills/init/init.ts
ls: cannot access '...': No such file or directory        # what the shim looks for

$ find ~/.claude/plugins/cache/roxabi-marketplace/dev-init -name init.ts -path '*skills/init*'
.../dev-init/900722eace17/skills/init/init.ts             # where it actually is
```

Observed with dev-core `0.8.16` + dev-init `900722eace17`.

## Workaround

Call dev-init directly, bypassing the shim:

```bash
[...]

Issue #348, opened 2026-07-16
fix(dev-core): ci-setup writes a partial dependabot.yml (github-actions only) that silently defeats Phase 1c's idempotency check


MickaelV0:
## Symptom

The `workflows` generator writes a `.github/dependabot.yml` containing **only** the `github-actions` block. Application dependencies are never tracked — the main reason to run Dependabot.

Verbatim output on `MickaelV0/demo-mcp`:

```yaml
version: 2
updates:
  - package-ecosystem: github-actions
    directory: /
    schedule:
      interval: weekly
      day: monday
    open-pull-requests-limit: 5
    cooldown:
      default-days: 3
      semver-major-days: 3
      semver-minor-days: 3
      semver-patch-days: 3
```

No `npm` / `pip` block. So GitHub Actions pins get updated; `hono`, `zod`, `react`, `wrangler`, `typescript` never do.

## Why it is a trap, not just an omission

`cookbooks/scanning.md` Phase 1c tells the agent to generate `.github/dependabot.yml` with **both** blocks:

```yaml
updates:
  - package-ecosystem: <ecosystem>     # ← from σ package_manager
    ...
  - package-ecosystem: github-actions
    ...
```

But Phase 1 (`workflows`) has already written a partial file by then. Since the skill is documented as **idempotent** ("skip already-configured items unless `--force`"), the natural reading at Phase 1c is *"dependabot.yml exists → skip"* — and the ecosystem block is never added. The two phases disagree about who owns the file, and the failure is silent: you end up with a valid, functioning Dependabot config that quietly covers half of what it should.

I only caught it by diffing the generated file against the cookbook's own YAML.

## Repro

```bash
bun <init.ts> workflows --stack bun --test none --deploy cloudflare \
  --merge merge-on-green --e2e none --lint false --typecheck true
cat .github/dependabot.yml     # github-actions block only
```

## Suggested fix

Pick one owner for the file:

- **Either** the generator emits both blocks — it already receives `--stack`, so the ecosystem is derivable there (`bun`/`npm`/`pnpm`/`yarn` → `npm`, `uv`/`pip` → `pip`);
- **or** the generator stops writing `dependabot.yml` entirely and leaves it to Phase 1c, which already has the full template and the σ `package_manager` value.

Whichever wins, Phase 1c's existence check should look *inside* the file for the ecosystem block rather than testing for the path — otherwise a partial file keeps defeating it.

## Note on the bun → npm mapping

[...]

Issue #347, opened 2026-07-16
fix(dev-core): ci-setup has no mapping for bun:test — bun repos get a green CI with no test step


MickaelV0:
## Symptom

`/ci-setup` gives a **bun** repo a `ci.yml` with **no test step at all**. CI goes green without ever running a test. A silently-green pipeline is worse than a red one: it certifies nothing while looking like it certifies everything.

## Root cause

Bun's built-in test runner (`bun test`) has no slot anywhere in the chain — even though `bun` is a first-class `runtime` everywhere else.

**1. `stack.yml.example` cannot express it:**

```yaml
testing:
  unit: vitest                    # vitest | jest | pytest | none
```

**2. `cookbooks/workflows.md` (Phase 1, step 3) has no mapping for it:**

> `test` ← `commands.test` (vitest→vitest, jest→jest, pytest→pytest, else→none)

So a repo whose `runtime: bun` and whose tests are `bun:test` falls into **`else` → `none`**.

**3. `--test none` then drops the step from the template.**

The generator offers `--stack bun` but no `--test bun`: the runtime is supported, its native test runner is not.

## Repro

`MickaelV0/demo-mcp` — `runtime: bun`, 5 passing `bun:test` tests in `packages/shared`:

```bash
$ bun run test
@demo-mcp/shared test: 5 pass
@demo-mcp/shared test: 0 fail          # tests exist and are green

$ bun <init.ts> workflows --stack bun --test none --lint false --typecheck true
```

produces:

```yaml
jobs:
  ci:
    steps:
      - uses: actions/checkout@…
      - uses: oven-sh/setup-bun@…
      - run: bun install
      - name: Typecheck
        run: bun typecheck
      # ← no Test step. CI is green. Nothing was tested.
```

Had to hand-add the step for CI to mean anything:

```yaml
      - name: Test
        run: bun run test
```

Verified green afterwards ([run](https://github.com/MickaelV0/demo-mcp/actions)) — the step works, it simply is never emitted.

## Suggested fix

- Add `bun` to `testing.unit` in `stack.yml.example` and to the `--test` enum.
- Map `bun` → a `bun run test` step in the ci.yml template (the command is already `commands.test` in σ — arguably the mapping should just *use* `commands.test` verbatim rather than re-deriving it from a runner enum).
- Consider making `--test none` emit an explicit comment in the generated `ci.yml` saying no test step was emitted and why. Right now the absence is silent and easy to miss on review.

[...]

Issue #346, opened 2026-07-16
fix(dev-core): env-setup scaffolds Critical Rules that duplicate the inherited parent CLAUDE.md/ssot context


MickaelV0:
## Symptom

`/env-setup` Phase 2 scaffolds a `## Critical Rules` block into the repo's `CLAUDE.md` without ever looking at the **inherited** context chain. For any repo under `~/projects/`, most of that block is already true one level up — so the scaffold writes a second copy of rules that are already in effect.

## Evidence

`~/projects/CLAUDE.md` `@import`s `ssot/operator.ssot.md`, whose line 17 reads:

> `work on issue → /dev #N = single entry (tier S/F-lite/F-full + lifecycle) | git: ¬--force ¬--hard ¬--amend ¬--no-verify | ∃ matching skill → use it (even w/o slash cmd)`

Mapped against what Phase 2 generated for `MickaelV0/demo-mcp`:

| Generated section | Already covered by | Verdict |
|---|---|---|
| 1. Dev Process (`/dev #N`, tiers S/F-lite/F-full) | `operator.ssot.md:17` | duplicate |
| 4. Git (`¬--force ¬--hard ¬--amend`) | `operator.ssot.md:17` | duplicate |
| Skills & Agents ("always use appropriate skill") | `operator.ssot.md:17` | duplicate |
| 2. Delegation · 3. Parallel · 5. Artifact Model · 6. Worktree · 7. Code Review | generic dev-core convention the skills carry themselves | boilerplate |
| TL;DR / Gotchas / Coding Standards | — | genuinely repo-local |

Net: **112 lines scaffolded, ~49 of them redundant.** Trimming to the repo-specific remainder left 63.

## Why this is more than noise

1. **Drift against the SSoT.** Two copies of "tiers = S/F-lite/F-full" means the day `operator.ssot.md` changes, every scaffolded repo silently contradicts it. `conventions.ssot.md` forbids exactly this (SSoT = one home per decision), and `/cleanup-context` exists to hunt it (`bloat=0`).
2. **Per-session cost.** `CLAUDE.md` loads on every session in that repo — the duplicate is paid every time, forever.
[...]

PR #354, opened 2026-07-16
fix(dev-core,dev-init): init shim, bun:test CI, dependabot, scaffold facts

MickaelV0:
## Summary

Closes the four demo-mcp `/env-setup` + `/ci-setup` defects, with panel-validated shapes.

| Issue | Fix |
|-------|-----|
| **#345** | `dev-core` init shim: flat sibling first, then versioned cache (`…/dev-init/<ver>/…`), skip `.orphaned_at` |
| **#347** | Parse `testing.unit`; classify `bun run test` ≠ none; emit `commands.test` in CI; `--test bun`; comment when no test step |
| **#348** | `generateDependabotYml(stack)` → ecosystem + github-actions; doctor content check; Phase 1c verifies/top-ups |
| **#346** | Scaffold facts from repo (base branch, PM, `.env.example`); parent CLAUDE.md report; Phase 2 user gate (no silent auto-skip) |

dev-core **0.8.17**.

## Test plan

- [x] Full vitest suite (578 pass)
- [x] pre-commit lint + typecheck + trufflehog
- [ ] Re-run `/env-setup` Phase 2 on a child of `~/projects/` — parent paths shown, facts correct
- [ ] `bun $I_TS workflows --stack bun --test bun --test-command "bun run test"` → CI has Test step
- [ ] Generated dependabot.yml has `npm` + `github-actions` (gha cooldown `default-days` only)

### DIFF
--- a/plugins/dev-core/.claude-plugin/plugin.json
+++ b/plugins/dev-core/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "dev-core",
-  "version": "0.8.16",
+  "version": "0.8.17",
   "description": "Full development workflow — frame, shape, plan, implement, review, ship. 30 skills, 9 agents, safety hooks.",
   "author": {
     "name": "Roxabi"
--- a/plugins/dev-core/hooks/lib/parse-stack-yml.cjs
+++ b/plugins/dev-core/hooks/lib/parse-stack-yml.cjs
@@ -174,22 +174,43 @@ function parseCommand(text, key) {
 }
 
 /**
- * Parse `testing.e2e` from stack.yml text.
+ * Parse a key under the `testing:` section from stack.yml text.
  *
  * @param {string} text
+ * @param {string} key
  * @returns {string|null}
  */
-function parseTestingE2e(text) {
+function parseTestingKey(text, key) {
   if (!text) return null
   const section = text.match(/^testing:\s*$/m)
   if (!section) return null
   const after = text.slice(section.index + 'testing:'.length)
   const nextTop = after.match(/\n\S/)
   const block = nextTop ? after.slice(0, nextTop.index) : after
-  const match = block.match(/^\s+e2e:\s*(\S+)/m)
+  const match = block.match(new RegExp(`^\\s+${key}:\\s*(\\S+)`, 'm'))
   if (!match) return null
   const val = match[1]
   return val === 'none' ? null : val
+}
+
+/**
+ * Parse `testing.e2e` from stack.yml text.
+ *
+ * @param {string} text
+ * @returns {string|null}
+ */
+function parseTestingE2e(text) {
+  return parseTestingKey(text, 'e2e')
+}
+
+/**
+ * Parse `testing.unit` from stack.yml text (vitest | jest | pytest | bun).
+ *
+ * @param {string} text
+ * @returns {string|null}
+ */
+function parseTestingUnit(text) {
+  return parseTestingKey(text, 'unit')
 }
 
 /**
@@ -259,6 +280,7 @@ function parseStandards(text) {
  *   standards: Record<string, string>|null,
  *   runtime: string|null,
  *   commands: {lint: string|null, typecheck: string|null, test: string|null},
+ *   testingUnit: string|null,
  *   testingE2e: string|null,
  *   ciMerge: string|null
  * }}
@@ -277,6 +299,7 @@ function parseStackYml(text) {
       typecheck: parseCommand(text, 'typecheck'),
       test: parseCommand(text, 'test'),
     },
+    testingUnit: parseTestingUnit(text),
     testingE2e: parseTestingE2e(text),
     ciMerge: parseCiMerge(text),
   }
@@ -292,6 +315,7 @@ module.exports = {
   parseStandards,
   parseRuntime,
   parseCommand,
+  parseTestingUnit,
   parseTestingE2e,
   parseCiMerge,
 }
--- a/plugins/dev-core/hooks/lib/parse-stack-yml.d.cts
+++ b/plugins/dev-core/hooks/lib/parse-stack-yml.d.cts
@@ -18,6 +18,7 @@ export interface StackYml {
   standards: Record<string, string> | null
   runtime: string | null
   commands: StackCommands
+  testingUnit: string | null
   testingE2e: string | null
   ciMerge: string | null
 }
@@ -31,5 +32,6 @@ export declare function parsePackageManager(text: string | null): string | null
 export declare function parseStandards(text: string | null): Record<string, string> | null
 export declare function parseRuntime(text: string | null): string | null
 export declare function parseCommand(text: string | null, key: string): string | null
+export declare function parseTestingUnit(text: string | null): string | null
 export declare function parseTestingE2e(text: string | null): string | null
 export declare function parseCiMerge(text: string | null): string | null
--- a/plugins/dev-core/skills/checkup/__tests__/dependabot-cooldown.test.ts
+++ b/plugins/dev-core/skills/checkup/__tests__/dependabot-cooldown.test.ts
@@ -1,0 +1,147 @@
+import * as fs from 'node:fs'
+import * as os from 'node:os'
+import * as path from 'node:path'
+import { afterEach, describe, expect, it } from 'vitest'
+import { checkSecurity, detectDependabotCooldownViolations } from '../doctor-local'
+
+/**
+ * Dependabot cooldown validity.
+ * `semver-*-days` is supported for npm/yarn but not for github-actions; GitHub
+ * rejects the whole file at parse time when it appears there, so a presence-only
+ * check passes on a config that is entirely inert.
+ */
+
+// The repo's own valid config, inlined so the test does not read .github/.
+const VALID = [
+  'version: 2',
+  'updates:',
+  '  - package-ecosystem: github-actions',
+  '    directory: /',
+  '    schedule:',
+  '      interval: weekly',
+  '    groups:',
+  '      github-actions:',
+  '        patterns:',
+  '          - "*"',
+  '        update-types:',
+  '          - minor',
+  '          - patch',
+  '    cooldown:',
+  '      default-days: 3',
+  '  - package-ecosystem: npm',
+  '    directory: /',
+  '    schedule:',
+  '      interval: weekly',
+  '    groups:',
+  '      npm-dependencies:',
+  '        patterns:',
+  '          - "*"',
+  '        update-types:',
+  '          - minor',
+  '          - patch',
+  '    cooldown:',
+  '      default-days: 3',
+  '      semver-major-days: 7',
+  '      semver-minor-days: 3',
+  '      semver-patch-days: 3',
+].join('\n')
+
+// The config GitHub rejected for weeks: semver-*-days under github-actions.
+const BROKEN = VALID.replace(
+  ['    cooldown:', '      default-days: 3', '  - package-ecosystem: npm'].join('\n'),
+  ['    cooldown:', '      default-days: 3', '      semver-major-days: 7', '  - package-ecosystem: npm'].join('\n'),
+)
+
+describe('detectDependabotCooldownViolations', () => {
+  it('does not flag the valid config, where only npm carries semver-*-days', () => {
+    expect(detectDependabotCooldownViolations(VALID)).toEqual([])
+  })
+
+  it('flags semver-major-days under github-actions', () => {
+    expect(detectDependabotCooldownViolations(BROKEN)).toEqual([
+      { ecosystem: 'github-actions', property: 'semver-major-days' },
+    ])
+  })
+
+  it('flags every offending key, not just the first', () => {
+    const yml = [
+      'updates:',
+      '  - package-ecosystem: github-actions',
+      '    cooldown:',
+      '      default-days: 3',
+      '      semver-major-days: 7',
+      '      semver-minor-days: 3',
+      '      semver-patch-days: 3',
+    ].join('\n')
+    expect(detectDependabotCooldownViolations(yml).map((v) => v.property)).toEqual([
+      'semver-major-days',
+      'semver-minor-days',
+      'semver-patch-days',
+    ])
+  })
+
+  it('ignores semver-*-days named in a comment inside a github-actions block', () => {
+    const yml = [
+      'updates:',
+      '  - package-ecosystem: github-actions',
+      '    # semver-major-days: 7 is rejected here — do not add it back',
+      '    cooldown:',
+      '      default-days: 3',
+    ].join('\n')
+    expect(detectDependabotCooldownViolations(yml)).toEqual([])
+  })
+
+  it('does not attribute a later npm key to an earlier github-actions block', () => {
+    const yml = [
+      'updates:',
+      '  - package-ecosystem: github-actions',
+      '    cooldown:',
+      '      default-days: 3',
+      '  - package-ecosystem: npm',
+      '    cooldown:',
+      '      semver-major-days: 7',
+    ].join('\n')
+    expect(detectDependabotCooldownViolations(yml)).toEqual([])
+  })
+
+  it('does not flag keys outside a cooldown mapping', () => {
+    const yml = ['updates:', '  - package-ecosystem: github-actions', '    groups:', '      semver-major-days: 7'].join(
+      '\n',
+    )
+    expect(detectDependabotCooldownViolations(yml)).toEqual([])
+  })
+})
+
+describe('checkSecurity dependabot.yml status', () => {
+  const original = process.cwd()
+  let tmpDir: string
+
+  function writeDependabot(content: string) {
+    tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'dependabot-test-'))
+    fs.mkdirSync(path.join(tmpDir, '.github'))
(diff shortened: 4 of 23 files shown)


MickaelV0:
## Code Review — PR #354

**Verdict: Request changes.** The four stated fixes are real, the generator fix is verified correct against GitHub's live schema, and the test suite claim ("578 pass") checks out. Three things block, and they compound in a way worth seeing together.

Reviewed by fresh agents (architect / tester / devops / security-auditor); every claim below was re-verified by executing something, not by reading.

---

### 🔴 Blocking

**1. `issue:` `/checkup` crashes at import in every cache install — so this PR's own #348 doctor check can never run in production.**
`plugins/dev-core/skills/checkup/workflow-drift.ts:11,17`

```
$ cd ~/.claude/plugins/cache/roxabi-marketplace/dev-core/0.8.16 && bun skills/checkup/doctor.ts
error: Cannot find module '../../../dev-init/skills/init/lib/workflows'
       from '.../dev-core/0.8.16/skills/checkup/workflow-drift.ts'
```

`../../../dev-init` resolves to `cache/roxabi-marketplace/dev-core/dev-init/` — **the version segment is missing**. That is exactly the #345 bug this PR fixes one directory over. ESM evaluates the whole module graph before any code runs, and `doctor.ts` imports both `checkWorkflowDrift` and `checkSecurity`, so the crash at import means **`checkSecurity` — the function carrying the new dependabot check — never executes**. It passes CI only because vitest runs from the repo, where flat siblings resolve.

Root cause: the PR internalized that the cache is version-scoped and fixed the *spawn-based* shim, which can glob. A **static ESM import cannot glob** — it is unfixable in place. This is the cross-plugin runtime import that CLAUDE.md forbids under marketplace self-containment; the PR extends the coupling without confronting it.

- Recommended: descope the real fix (copy-sync `workflows*.ts` per `tools/shared-sources.json`) to a follow-up — **it inverts the canonical direction** (all 14 current entries are dev-core-canonical; these generators live in dev-init), which is an ownership decision, not a mechanical change. Here, take the ~5-line unblock: wrap the `checkWorkflowDrift()` call behind a dynamic `await import()` in try/catch so a resolution failure degrades that one check to `warn` instead of taking down all of `/checkup`.

**2. `issue:` `writeWorkflows` clobbers `dependabot.yml` without `--force` and doesn't report it — it would silently revert #350 *and* #342 on this repo.**
`plugins/dev-init/skills/init/lib/workflows.ts` (`writeWorkflows(opts: WorkflowOpts)` — **no `force` parameter**; unconditional `fs.writeFileSync('.github/dependabot.yml', …)`; no `written.push('dependabot.yml')`)

The cookbooks promise `--force # only when topping up a partial dependabot.yml` and *"les fichiers déjà présents sur le repo sont skippés."* Both are false for `writeWorkflows` — only `pushWorkflows` honours `skipExisting: !force`. Verified against a temp repo seeded with this repo's real post-#350 file: clobbered without `--force`, `semver-major-days:7` occurrences → **0**. The file is written but omitted from the `written[]` report, so the JSON under-reports what it touched.

**This compounds with finding 4:** the same run also rewrites `auto-merge.yml` from the generator — which still ships the dead title-regex guard. So one `/ci-setup` on this repo reverts **both** #350's cooldown fix and #342's `fetch-metadata` guard, silently, in one shot.

- Recommended: thread `force` into `writeWorkflows(opts, force = false)`; `if (existsSync(p) && !force) → skip + report`, else write + `written.push(...)`. Makes the cookbook true.

**3. `suggestion(blocking):` The new dependabot doctor check validates *presence*, not *validity* — it reports ✅ `pass` on the exact config GitHub silently rejected here for weeks.**
`plugins/dev-core/skills/checkup/doctor-local.ts:114-121`

Ran the PR's exact regexes against the real pre-#350 file (`git show 57d1948^:.github/dependabot.yml`):

```
rejected.yml   hasGha=true hasApp=true → doctor: PASS ✅ "found (ecosystem + github-actions)"
valid.yml      hasGha=true hasApp=true → doctor: PASS ✅ "found (ecosystem + github-actions)"
```

The gate cannot tell the parse-rejected config from the working one. Presence and validity are different properties; a gate asserting the first while its name implies the second turns a silent failure into a **certified-healthy** one.

**Two corrections that matter for the fix:**
- The blind population is *not* "projects from the old generator" — those have only the github-actions block → `hasApp=false` → correctly `warn`. It is projects built from the **old Phase 1c cookbook**, which taught both blocks *and* the illegal `semver-*-days` (this PR deletes exactly those four lines at `scanning.md:105-108`). That population includes **roxabi-plugins itself, pre-#350**.
- **The obvious fix inverts the bug.** A whole-file `/semver-major-days/` test false-positives on the *current valid* file — npm legitimately carries `semver-major-days: 7` (added in #342). Verified: naive whole-file → FLAGS on both. It **must be block-scoped**:

```ts
const blocks = depYml.split(/\n(?=\s*- package-ecosystem:)/)
const ghaBlock = blocks.find(b => /package-ecosystem:\s*github-actions/.test(b))
const illegal = ghaBlock && /semver-(major|minor|patch)-days/.test(ghaBlock)
// → status 'fail': GitHub rejects the ENTIRE file, both ecosystems, silently (see #350)
```

Status `fail`, not `warn`. Note the identical presence-only contract is implemented **twice with no shared source** — TS at `doctor-local.ts:114-115` and bash at `scanning.md:69-72` (`parallel-path-drift`); any fix must land in both.

Scope-wise this is a defensible follow-up — #348 only asks Phase 1c to "look inside the file for the ecosystem block," and the old `existsSync` check passed on the rejected file too, so it is **not a regression**. Recommending it here anyway: this is the dependabot-correctness PR, it is ~6 lines, and the alternative ships a green light over a trap that burned this repo last week.

**4. `issue:` [`test-tautology`] The `.orphaned_at` skip guard — #345's stated deliverable — has no test that fails when the guard is deleted.**
`plugins/dev-core/skills/init/__tests__/init-shim.test.ts:46-57`

Falsified: removed `if (existsSync(join(versionDir, '.orphaned_at'))) continue` from `init.ts:33` → **all 3 tests still pass**. Restored → green. The test passes for an unrelated reason: `.orphaned_at` is written into `old` *after* creation, bumping its mtime, and `new` is created after that — so `new` wins on mtime alone whether or not the filter runs.

Same blind spot one level up: the newest-mtime tie-break (`init.ts:42`) is also unexercised — replacing it with "keep first found" broke nothing. No test ever has two *non-orphaned* valid version dirs, which is the realistic post-upgrade cache shape #345 was filed against.

- Recommended: one fixture with two valid dirs + explicit `utimesSync` covers both the `.orphaned_at` exclusion and the tie-break.

**5. `todo:` `--help` still writes 7 files, while the PR says "Closes #345".**
`plugins/dev-init/skills/init/init.ts`

```
$ bun plugins/dev-init/skills/init/init.ts workflows --help
→ writes .github/dependabot.yml + 6 workflow files
```

#345 raises this explicitly as its secondary defect. `hasFlag` already exists and the usage text already exists as the file docstring — ~4 lines. Otherwise drop `#345` from the Closes list so the issue isn't auto-closed while live.

---

### 🟡 Non-blocking, but worth knowing

- `issue:` **The generator still ships the dead title-regex guard to every scaffolded project.** `workflows.ts:70` carries the exact `PR_TITLE =~ from\ v?([0-9]+)…` regex proven inert in #342, while the new `generateDependabotAutomergeYml` correctly uses `fetch-metadata` (`workflows-fleet.ts:82-89`). Today the two agree only *by accident* — the group's `update-types: [minor,patch]` structurally keeps majors out, so majors arrive as individual PRs whose titles the regex happens to parse. Two independent gates agreeing by coincidence of config, not construction. Recommend removing the regex step, or rewriting it on `fetch-metadata` so both agree by design.
- `issue:` **`bun` has been its own Dependabot ecosystem since Feb 2025** (GA, Bun ≥ v1.1.39) — `dependabotEcosystemFromStack` maps `bun → npm`. The cookbook itself hedges: *"bun → Dependabot `npm` ecosystem (fleet convention). Lockfile support for `bun.lock` is GitHub-side — **not re-verified here**"* (`scanning.md:118`). That hedge is the same unverified-per-ecosystem assumption that shipped the `semver-*-days` bug. (This repo carries **both** `bun.lock` and `package-lock.json`, so `npm` works *here* — the exposure is for bun-only projects.) Either switch to `bun` after verifying it actually updates `bun.lock`, or replace the hedge with a dated, linked verification note.
- `issue:` `missing-input-validation` — `--test-command` / `commands.test` is spliced **unquoted** into the generated CI `run:` line (`workflows.ts:334`). Benign and likely: `commands.test: bun test --grep "foo: bar"` emits a plain scalar containing `: ` → the pushed `ci.yml` **fails to parse**. Hostile and conditional: a newline in the flag appends arbitrary steps (argv carries newlines; stack.yml cannot, so the flag is the vector — operator-local, not fork-reachable). Fix: emit a block scalar (`run: |` + indent). Note the current test only does `toContain`, which a malformed doc passes.
- `nitpick:` **`ACTION_PINS.setupNode` is 39 hex chars — not a valid SHA-1.** `actions/setup-node@49933ea53b297553b1c3d69c20d3553fa2791a6` → `gh api` returns `422 No commit found for SHA`. All 7 sibling pins are exactly 40. Real `v4` is `49933ea5288caeca8642d1e84afbd3f7d6820020` (shares the `49933ea5` prefix — a corrupted transcription of exactly this). **Every generated node-stack CI is dead on arrival.** Fails closed, so correctness not supply-chain. **Pre-existing — `workflow-pins.ts` is not in this PR.** Worth its own PR + a test asserting `/^[0-9a-f]{40}$/` on every pin.
- `suggestion(non-blocking):` Coverage gaps on this PR's own deliverables: `detectBaseBranch` (#346's core "derive facts, don't assume" fix — **zero** tests; the exact regression #346 reported is unverified), `filterProjectLocalSections` (#346's Phase-2 gate option — no test imports it), and the new doctor dependabot check (no test on any of its three branches).
- `nitpick:` `resolveTestRunCommand` has three conditionals whose branches all return the same value (`workflows.ts:294,297-302`), and `classifyTestRunner`'s `/^vitest$/ || /vitest/` pairs are subsumption no-ops (`:591-593`). Post-#347 the runner identity is decorative — only `none` vs not-`none` matters, plus `testCommand`. Worth stating rather than disguising as live branching. (`/^bun$/ || /bun:test/` at `:594` is genuinely not redundant — leave it.)
- `suggestion(non-blocking):` `generateDependabotYml`'s `{stack} | {ecosystem}` union has **zero** callers on the `{ecosystem}` arm, and `stack: WorkflowOpts['stack'] | string` deletes the exhaustiveness that makes the mapping provably total over the closed `'bun'|'node'|'python'` union. Drop both.

---

### ✅ Verified clean / praise

- `praise:` **`generateDependabotYml` is correct against the live schema and genuinely tested.** Independently confirmed per-block: both ecosystems emitted, github-actions cooldown reduced to `default-days` only. The pattern-less `groups.minor-and-patch` with bare `[minor, patch]` is **legal** (that's the right form; the `version-update:semver-*` prefix belongs to top-level `ignore`, not groups). And the regression guard is real, not decorative — reintroducing `semver-major-days: 3` into the gha block **fails the test**. The parallel deletion in `scanning.md:105-108` is the edit that actually stops propagation, and it would have been easy to miss.
- `praise:` **Init shim path resolution is safe.** `<ver>` comes from `readdirSync`, which returns basenames only and never `.`/`..` — traversal is unrepresentable by construction, not merely unvalidated. `Bun.spawnSync` is array-form (no shell). `import.meta.main` guard is the right call.
- `praise:` **No secret leak in the scaffold facts.** `.env.example` is touched by `existsSync` only — its **values are never read**; the sole output is a boolean. A real token pasted into a project's `.env.example` cannot reach the report. `detectPackageInstall`/`detectBaseBranch` map through closed sets, so config→generated-file flow is whitelisted, not reflected.
- `praise:` No PR-controlled data reaches a `run:` block; all PR-derived values go through `env:` and are referenced as shell vars — consistently, including in the new `dependabot-automerge.yml`. No `pull_request_target` in any generator.
- `praise:` Version bump `0.8.16 → 0.8.17` correct on both halves — dev-core is semver-disciplined and bumps; `marketplace.json` carries no `version`, so the "never in both" trap is avoided.
- `thought:` **Correction to a prior review:** `plugins/dev-core/skills/init/init.ts` is **not** dead code. It has no SKILL.md (dev-core exposes no `init` skill — correct), but it is a **CLI shim** that three live dev-core skills resolve to as `$I_TS` (`ci-setup:12`, `env-setup:12`, `checkup:15`). That is precisely why #345 killed `/env-setup` Phase 2. This PR revives genuinely live code and its test targets a real entry point. (The tsconfig `exclude` of that file is now ineffective — the new test imports it, so TS follows it regardless — which is an improvement, but makes the `exclude` entry misleading.)

MickaelV0:
## Blocking findings addressed — `a01cd6b`

All five blockers from the review above are fixed. Each was verified **by execution against a real cache layout**, not by reading. Gates: `typecheck` clean, `lint` 0 new warnings (28 pre-existing, identical before/after), **592 passing** (was 578, +14), `validate_plugins.py` all green.

### 1. `/checkup` crashed at import in every cache install (critical)

`workflow-drift.ts` imported dev-init's generators through a **static relative path**. From `…/cache/<mp>/dev-core/<ver>/skills/checkup/`, `../../../` lands on `dev-core/dev-init/` — the version segment is missing. And dev-init's cache dir is a **commit SHA** (`900722eace17`), so **no static path can ever reach it**.

`doctor.ts` imports the drift check eagerly, so ESM evaluated the whole graph and **all of `/checkup` died** — including the dependabot check this PR adds for #348. The PR shipped an unreachable safeguard. It passed CI only because vitest runs from the repo, where flat siblings resolve.

Generators are now resolved and imported **at call time** via the shim's own proven resolver (split into a non-exiting `tryResolveDevInitEntry`; the CLI's exit-1 contract is unchanged). Absent dev-init degrades to `skip` rather than throwing — dev-core must not hard-depend on dev-init at module load.

Verified in a reconstructed SHA-versioned cache: previously `Cannot find module '../../../dev-init/…'`; now returns real drift results. With dev-init deleted: `status: skip`, no throw.

### 2. `writeWorkflows()` clobbered everything, silently

It overwrote every file unconditionally, while its siblings `pushWorkflows`/`pushGenericWorkflows` both document **top-up by default, `force` overwrites**. It also wrote `.github/dependabot.yml` with no entry in its report: measured **7 files written, 6 reported**.

One local run **reverted #342's `fetch-metadata` guard back to the dead `BASH_REMATCH` regex** — silently, and invisibly in the report.

Both paths now derive from one `workflowFileSet()`, so they cannot diverge again, and the local writer honours `force = false`. Verified with sentinels: default run leaves them intact and reports `dependabot.yml: skipped`; `--force` reports `updated` and overwrites.

### 3. `workflows` local-wrote on any missing flag

`if (owner && repo)` meant a typo — or `--help` — fell through to the local write and destroyed files. `bun init.ts workflows --help` wrote 7 files in any repo with a git remote. Local write is now explicit; `--help` prints usage.

### 4. The dependabot check validated presence, not validity

It grepped for ecosystem *presence* and reported `pass` on the exact config GitHub rejected wholesale for weeks — it could not detect the bug it was created for.

**The naive fix inverts it:** a whole-file `/semver-major-days/` match flags **this repo's currently valid config**, because npm legitimately carries `semver-major-days: 7`. The check now attributes each key to its containing `- package-ecosystem:` block. Hand-parsed — no YAML dep available.

| fixture | result |
|---|---|
| this repo's real `dependabot.yml` (valid) | `[]` — not flagged |
| historical broken config | flags `semver-major-days` + `semver-minor-days` under `github-actions`, ignores npm's legal one |

### 5. The `.orphaned_at` test was tautological

Mutation-tested: with `if (existsSync(join(versionDir, '.orphaned_at'))) continue` **deleted, all 3 tests still passed** — the `good` dir won on mtime regardless. The orphaned dir is now explicitly the newest (fixed epochs via `utimesSync`, no creation-order race), so only the guard can exclude it. Re-mutated to prove it now **fails** without the guard, then restored.

---

### Correction to my own verdict

I wrote that regenerating would revert **both #350 and #342**. That overstated it, and I should have checked before saying it. The generator emits only `default-days: 3` under `github-actions`, so it does **not** re-introduce the illegal properties — config validation stays intact. It reverts **#342's guard** and loses tuning (the github-actions `update-types` split, npm's cooldown). #350 is not at risk.

### Left deliberately out of scope

- **Cross-plugin runtime import.** CLAUDE.md forbids these under marketplace self-containment; the doctrinally pure fix is copy-sync. Making the import dynamic makes it *work*, not *compliant*. `tools/shared-sources.json` is uniformly dev-core-canonical → dev-init, whereas these generators are dev-init-canonical under `skills/init/lib/` — adopting them **inverts the manifest's ownership direction**. That's a design decision, not a review fix.
- **`ACTION_PINS.setupNode` is a nonexistent commit** — 39 chars, `gh api` → `422 No commit found`; real v4 is `49933ea5288caeca8642d1e84afbd3f7d6820020`. Every other pin is 40. Any **node-stack** repo scaffolded by dev-init gets a CI that dies at "Set up job". This repo is bun-stack, so its own CI is unaffected. The file isn't in this PR — needs its own.
- `SEMVER_COOLDOWN_UNSUPPORTED` covers only `github-actions`; docker/gomod/cargo may share the class. Widening is one line, but I did not verify their support-table rows, and a miss fails safe where a false positive would not.

PR #357, opened 2026-07-17
fix(dev-init): generate the fetch-metadata semver-major guard, not the dead regex

MickaelV0:
## Problem

Every project scaffolded by `/init` or `/ci-setup` shipped a **broken** dependabot auto-merge guard. `generateAutoMergeYml()` emitted a PR-title regex:

```bash
if [[ "$PR_TITLE" =~ from\ v?([0-9]+)[^\ ]*\ to\ v?([0-9]+) ]] && [ "${BASH_REMATCH[1]}" != "${BASH_REMATCH[2]}" ]; then
```

It fails in **both** directions:
- **Never fires on grouped PRs** — Dependabot titles them `bump the <group> group…` with no versions, so a major hidden inside a group merges automatically.
- **Can false-fire on SHA-pinned action bumps** — `bump actions/checkout from 08eba0b to 8f4b7f8` — digit-leading SHAs match the regex and a patch bump gets refused.

This repo's own `auto-merge.yml` replaced it with `dependabot/fetch-metadata` in #342. The **generator** kept emitting the regex — so `/checkup` now (correctly) reports `drift:auto-merge.yml`, and every fresh scaffold inherits the bug.

## Fix

Port #342's approach into the generator: read `steps.dependabot-meta.outputs.update-type` and block on `version-update:semver-major`. For a grouped PR, `fetch-metadata` reports the **highest** update-type in the group, so a major hidden in a group is caught.

- Generated YAML **parse-checked** (both `auto-merge.yml` and `dependabot-automerge.yml`, `id`s intact).
- **Regression test** asserts `BASH_REMATCH`/`PR_TITLE` cannot return and that the block gates on the metadata output. Mutation-tested: re-inlining the regex fails the test.

## Governance gap the advisory surfaced

`fetch-metadata` was **hardcoded inline** at `v2.5.0` in `generateDependabotAutomergeYml`, bypassing `ACTION_PINS` — and therefore bypassing the `verify:pins` guard from #355. It is now an `ACTION_PINS` entry (`v3.1.0`, matching #342), referenced from both generators.

`tools/verify-action-pins.ts` is extended so this class cannot recur: it now scans the emitter sources for `uses: owner/repo@sha` literals and **fails any SHA not declared in `ACTION_PINS`**. Mutation-tested against re-inlining `v2.5.0`:

```
MUTANT (inline v2.5.0)  -> exit 1  FAIL …/workflows-fleet.ts: uses dependabot/fetch-metadata@21025c705c08 — not declared in ACTION_PINS
RESTORED                -> exit 0  All 9 pins resolve; no inline pin bypasses ACTION_PINS.
```

## Trap checked, not assumed

`fetch-metadata` `v3.0.0`'s only breaking change is requiring **node24 on the runner** — `ubuntu-latest` supplies it. The `update-type` outputs are unchanged from `v2`, and the SHA (`25dd0e34…`) is exactly this repo's `v3.1.0`, which blocked #306 in production. No new secret/permission: the generated workflow already mints the App token (`steps.app.outputs.token`).

## Gates

`lint` 0 errors, `typecheck` clean, **593 tests** pass, `validate_plugins.py` green, `verify:pins` green (9 pins, no bypass).

## Scope

Advisory Q2. Q1 (the cross-plugin ownership decision) and Q3 (widening the cooldown-unsupported ecosystem set, pending a doc re-check) are deliberately separate. No dependency on the open PRs.

### DIFF
--- a/plugins/dev-init/skills/init/__tests__/workflows.test.ts
+++ b/plugins/dev-init/skills/init/__tests__/workflows.test.ts
@@ -42,6 +42,16 @@ describe('generateAutoMergeYml', () => {
     expect(yml).toContain(ACTION_PINS.githubScript)
     expect(yml).not.toContain('actions/github-script@v8')
   })
+
+  it('blocks semver-major via fetch-metadata, not the dead title regex', () => {
+    const yml = generateAutoMergeYml()
+    // The title regex never fired on grouped PRs (no versions in the title) and
+    // could misread SHA-pinned action bumps — #342 replaced it with metadata.
+    expect(yml).not.toContain('BASH_REMATCH')
+    expect(yml).not.toContain('PR_TITLE')
+    expect(yml).toContain(ACTION_PINS.dependabotFetchMetadata)
+    expect(yml).toContain("steps.dependabot-meta.outputs.update-type == 'version-update:semver-major'")
+  })
 })
 
 describe('generateCiYml', () => {
--- a/plugins/dev-init/skills/init/__tests__/workflows.test.ts
+++ b/plugins/dev-init/skills/init/__tests__/workflows.test.ts
@@ -50,7 +50,23 @@ describe('generateAutoMergeYml', () => {
     expect(yml).not.toContain('BASH_REMATCH')
     expect(yml).not.toContain('PR_TITLE')
     expect(yml).toContain(ACTION_PINS.dependabotFetchMetadata)
-    expect(yml).toContain("steps.dependabot-meta.outputs.update-type == 'version-update:semver-major'")
+
+    // Derive the reference from the declared id — a rename of `id:` that forgets
+    // to update the block's `if:` must fail this, not just an equality check.
+    const fetchIdMatch = yml.match(/- name: Fetch dependabot metadata\s*\n\s*id: (\S+)/)
+    if (!fetchIdMatch) throw new Error('Fetch dependabot metadata step id not found')
+    const fetchId = fetchIdMatch[1]
+
+    const blockStart = yml.indexOf('- name: Block dependabot semver-major')
+    expect(blockStart).toBeGreaterThan(-1)
+    const nextStepOffset = yml.slice(blockStart + 1).search(/\n\s*- name: /)
+    const blockRegion =
+      nextStepOffset === -1 ? yml.slice(blockStart) : yml.slice(blockStart, blockStart + 1 + nextStepOffset)
+
+    expect(blockRegion).toContain(`steps.${fetchId}.outputs.update-type == 'version-update:semver-major'`)
+    // Scoped to the Block step only — the whole YAML also has a legitimate
+    // `exit 0` elsewhere (update-behind-prs' empty-PR-list check).
+    expect(blockRegion).toContain('exit 1')
   })
 })
 
@@ -150,6 +166,8 @@ describe('generateDependabotAutomergeYml', () => {
     expect(yml).toContain('dependabot[bot]')
     expect(yml).toContain('semver-patch')
     expect(yml).toContain(ACTION_PINS.createAppToken)
+    expect(yml).toContain(ACTION_PINS.dependabotFetchMetadata)
+    expect(yml).not.toContain('21025c705c08')
   })
 })
 
--- a/plugins/dev-init/skills/init/lib/workflow-pins.ts
+++ b/plugins/dev-init/skills/init/lib/workflow-pins.ts
@@ -11,4 +11,5 @@ export const ACTION_PINS = {
   githubScript: 'actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3', // v9
   semanticPr: 'amannn/action-semantic-pull-request@48f256284bd46cdaab1048c3721360e808335d50', // v6
   createAppToken: 'actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1', // v3.2.0
+  dependabotFetchMetadata: 'dependabot/fetch-metadata@25dd0e34f4fe68f24cc83900b1fe3fe149efef98', // v3.1.0
 } as const
--- a/plugins/dev-init/skills/init/lib/workflows-fleet.ts
+++ b/plugins/dev-init/skills/init/lib/workflows-fleet.ts
@@ -106,7 +106,7 @@ ${APP_MINT_STEP}
 
       - name: Fetch dependabot metadata
         id: meta
-        uses: dependabot/fetch-metadata@21025c705c08248db411dc16f3619e6b5f9ea21a  # v2.5.0
+        uses: ${ACTION_PINS.dependabotFetchMetadata}
         with:
           github-token: \${{ secrets.GITHUB_TOKEN }}
 
--- a/plugins/dev-init/skills/init/lib/workflows.ts
+++ b/plugins/dev-init/skills/init/lib/workflows.ts
@@ -60,19 +60,30 @@ jobs:
     steps:
 ${APP_MINT_STEP}
 
-      - name: Block dependabot semver-major bumps
+      # Read the real update-type from Dependabot's metadata rather than parsing the
+      # PR title: grouped PRs are titled "bump the <group> group…" with no versions,
+      # so a title regex never fires for a major hidden in a group, and it can also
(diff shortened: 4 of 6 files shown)


MickaelV0:
## Code Review — PR #357

Fresh multi-domain review (security-auditor, devops, tester), every load-bearing claim re-verified by execution before posting.

**Verdict: 🔴 Request changes** — the *fix* is correct, but its *tests are tautological* and the pin guard has two bypass holes. Given this PR is itself about closing "check reports pass without catching the bug" gaps, these can't ride along.

### What is solid (verified, not asserted)

- **Generated `auto-merge.yml` is byte-for-byte the merged reference** (`.github/workflows/auto-merge.yml`) apart from prose + the EN/FR refusal message. Step ids, `if:` gates, `permissions:`, token usage identical. — devops
- **fetch-metadata v2.5.0 → v3.1.0 is output-compatible.** `action.yml` `outputs:` schema byte-identical; only `runs:` changed `node20 → node24`. Every generated workflow hardcodes `runs-on: ubuntu-latest`, so node24 is moot for all consumers. Pinned SHA `25dd0e34…` confirmed = v3.1.0. — devops
- **`EMITTER_PATHS` is complete** — repo-wide grep finds inline `uses:@sha` literals only in the 4 listed files. — devops
- **Security-positive.** Removing `PR_TITLE` from the guard's env eliminates the last GitHub-controlled free-text value from a `run:` block. `pull_request_target` path does no checkout, minimal permissions, actor-gated. `verify-action-pins.ts` `gh api` call is argv-array (no shell) → injection-safe. — security-auditor

### 🔴 Blockers

**1. `issue(blocking)`: the regression test pins strings, not behavior — two silent-disable mutations pass.**
`plugins/dev-init/skills/init/__tests__/workflows.test.ts:46`
Class: `test-tautology`
The test asserts `BASH_REMATCH`/`PR_TITLE` absent and the pin/condition substrings present. It proves the dead regex is *gone* — not that the new guard *works*. Mutation-verified (both re-run by the orchestrator, tree restored):
- rename `id: dependabot-meta` → the `if:` reference `steps.dependabot-meta.outputs.update-type` now resolves to empty, guard **never fires** → **32/32 still pass**.
- `exit 1` → `exit 0`, guard no longer fails the job → **32/32 still pass**.
This is the exact "presence ≠ validity" hollowness the doctor-check finding (#354) called out.
→ **Fix:** parse the generated YAML and assert the wiring — the step carrying `id: dependabot-meta` exists, the Block step's `if:` references *that same id*, and its `run:` ends in `exit 1`. Derive the expected reference from the declared id so a rename can't pass. — tester (C90)

**2. `issue(blocking)`: `verify-action-pins.ts` governs by SHA alone, not the (action, SHA) pair.**
`tools/verify-action-pins.ts:119`
`governed = new Set(pins.map(p => p.sha))` discards `p.action`. Mutation-verified: `uses: WRONG-ORG/wrong-action@25dd0e34…` (fetch-metadata's SHA under a bogus action) → **guard prints "no inline pin bypasses ACTION_PINS", exit 0**. A guard whose stated job is "a SHA that doesn't belong where it's used" accepts exactly that.
→ **Fix:** key `governed` by `${action}@${sha}` and compare `${inline.action}@${inline.sha}`. — devops (C85)

### 🟡 Warnings

**3. `suggestion(blocking)`: a floating-tag inline `uses: foo/bar@v3` evades the guard entirely** — two agents converged.
`tools/verify-action-pins.ts:63`
The regex requires `@[0-9a-fA-F]{7,}`, so a non-SHA ref (`@v3`, `@v3.1.0`) matches nothing: not flagged ungoverned, not resolved, invisible — while the banner still claims "no inline pin bypasses." An unpinned action is a *worse* supply-chain state than an ungoverned-but-pinned SHA. Either broaden the scan to flag any emitter `uses:@<non-SHA>`, or narrow the docstring/banner to the guarantee actually provided. — security-auditor (C66) + devops (C80)

**4. `issue(blocking)`: the fleet pin swap has no unit test.**
`plugins/dev-init/skills/init/__tests__/workflows.test.ts:148` / `workflows-fleet.ts:109`
Class: `test-tautology`
Mutation-verified: revert `generateDependabotAutomergeYml`'s pin to the old inline `@21025c7…  # v2.5.0` → **32/32 still pass**. `verify:pins` catches it in CI (network job), but the unit layer — which *does* have a regression test for the sibling `workflows.ts` change — has a hole for the very generator this PR touches.
→ **Fix:** one line — `expect(yml).toContain(ACTION_PINS.dependabotFetchMetadata)` in the existing `it` block. — tester (C85) + devops (C90)

### 🟢 Suggestion (non-blocking)

**5. `suggestion(non-blocking)`: `verify-action-pins.ts` pure parsing has zero test coverage.**
`tools/verify-action-pins.ts:50`
`findInlinePins`/`parsePins` are pure string logic but can't be imported without triggering the network SHA-resolution + `process.exit`. Sibling `plugins/dev-core/tools/licenseChecker.ts` sets the precedent: gate the top-level call behind `if (import.meta.main)`, export the pure functions, unit-test them offline. — tester (C65)

---

### Summary

| # | Finding | Class | Blocks | C |
|---|---------|-------|:---:|---|
| 1 | Regression test pins strings, not step wiring | test-tautology | ✓ | 90 |
| 2 | Guard governs SHA alone, not (action,SHA) | — | ✓ | 85 |
| 3 | Floating-tag inline `uses:` evades guard | — | ✓ | 80 |
| 4 | Fleet pin swap has no unit test | test-tautology | ✓ | 90 |
| 5 | Pure parsing funcs untestable (no `import.meta.main`) | — | ✗ | 65 |

The generated workflow is correct and verified. All five findings are cheap, and 1–4 are instances of the exact hollow-check pattern this PR series exists to eliminate — so fixing them here keeps the series honest rather than shipping a guard with the holes it was built to catch.

*Reviewed by fresh agents; claims re-verified by execution. Findings only — no code changed.*

MickaelV0:
## Review Fixes Applied — `d13cd1f`

All 5 findings from the review above fixed, by two fresh fixers (file-disjoint). **Every fix was accepted only after its mutation was confirmed to fail the check** (falsification gate, re-run by the orchestrator, not trusted from the agents).

**Applied:** 5 · **Deferred:** 0 · **Skipped:** 0 · **Failed:** 0

### Falsification gate — 5/5 mutations caught, 0 hollow

| Finding | Mutation injected | Result |
|---|---|---|
| F1 wiring | rename `id: dependabot-meta` (silently disables guard) | test now **fails** |
| F1 wiring | `exit 1` → `exit 0` (guard stops blocking) | test now **fails** |
| F4 fleet pin | revert pin to inline `@21025c7… v2.5.0` | test now **fails** |
| F2 governance | governed SHA under a **wrong action** | guard now **fails** |
| F3 governance | floating tag `@v3` (unpinned) | guard now **fails** |

Before the fixes, each of these passed green — the exact "reports pass without catching the bug" hollowness this PR series is about.

### What changed

**F1 — `workflows.test.ts`:** the regression test now extracts the Fetch step's declared `id` and asserts the Block step's `if:` references *that same id* (a rename that forgets the `if:` fails), and asserts the Block region ends in `exit 1` (scoped — the YAML has a legitimate `exit 0` elsewhere).

**F2 + F3 — `tools/verify-action-pins.ts`:** governance now matches the **(action, ref) pair**, not the SHA alone. `findInlinePins` captures any ref (not just hex), so a floating tag is caught instead of being invisible. One inline `uses:` whose pair isn't in `ACTION_PINS` → fail. The docstring/banner claim is now honest.

**F4 — `workflows.test.ts`:** added `toContain(ACTION_PINS.dependabotFetchMetadata)` + `not.toContain('21025c705c08')` to the fleet generator's test.

**F5 — `tools/verify-action-pins.ts` + new `tools/__tests__/verify-action-pins.test.ts`:** pure parse/governance helpers exported behind `if (import.meta.main)` (licenseChecker.ts precedent), covered by an offline unit test (14 cases). `import.meta.dir` → `import.meta.dirname ?? '.'` so the module imports under vitest.

### Gates
`lint` 0 errors (28 pre-existing warnings, unchanged), `typecheck` clean, **607 tests** (+14), `validate_plugins.py` green, `verify:pins` green (9 pins, no bypass).

*Fixes applied by fresh agents; the falsification gate was re-run by the orchestrator. `reviewed` label deliberately not set — merge is the author's call.*

Issue #1, opened 2026-03-01
Make dev-core plugins generic / project-agnostic


MickaelV0:
## Problem

The dev-core plugins (plan, implement, review, test, validate, etc.) are currently tied to specific project conventions (Bun, TurboRepo, Biome, NestJS, TanStack, etc.). They should be refactored to be fully generic and project-agnostic.

## Goal

Each dev-core plugin should auto-discover project structure, tooling, and conventions rather than assuming a specific stack or layout. This aligns with the roxabi-plugins design principle: **project-agnostic — auto-discover structure, don't assume layout**.

## Scope

- All dev-core skills (`/plan`, `/implement`, `/review`, `/test`, `/validate`, `/analyze`, `/spec`, etc.)
- All dev-core agents (frontend-dev, backend-dev, devops, tester, etc.)
- Agent tool descriptions and available tool sets
- Hardcoded references to specific frameworks/tools (Vitest, Playwright, Drizzle, Fastify, etc.)

## Approach

- Replace hardcoded stack references with auto-discovery logic (read package.json, CLAUDE.md, config files)
- Make agent descriptions generic (e.g., "frontend-dev" should work with any frontend framework, not just TanStack)
- Parameterize tooling references so they adapt to the project context
- Keep the ability to specialize when project context is detected, but don't require it

## Acceptance Criteria

- [ ] Plugins work correctly on a fresh project with no specific framework
- [ ] Plugins still work well on projects that match the current stack assumptions
- [ ] No hardcoded framework/tool names in skill descriptions or agent prompts (unless as examples)
- [ ] Auto-discovery reads project config files to determine tooling

PR #358, opened 2026-07-17
fix(checkup): widen dependabot cooldown check to all 12 default-days-only ecosystems

MickaelV0:
## What

Widen the `/checkup` dependabot-cooldown validity check from **1** ecosystem (`github-actions`) to the full **12** that accept `default-days` only.

GitHub rejects the *entire* `dependabot.yml` at parse time whenever `semver-*-days` appears under any ecosystem that supports `default-days` only — not just `github-actions`. The other eleven were unguarded: a config that GitHub silently ignores wholesale would pass `/checkup`. This is the same "presence ≠ validity" class the check was created (#348/#354) to close, applied to its own coverage gap.

## The 12 (verified against the live docs, 2026-07-17)

`bazel · devcontainers · docker · docker-compose · github-actions · gitsubmodule · helm · nix · opentofu · pre-commit · terraform · vcpkg`

Source: https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference

The complement — `npm, gomod, cargo, pip, uv, bun, bundler, composer, maven, gradle, nuget, hex, pub, swift, deno, dotnet, elm, conda, julia, sbt, rust` — **does** support `semver-*-days`. Flagging any of those is the exact false positive this check must never produce, so they stay out.

### One correction caught in verification
The earlier scratchpad guess for Nix was `nix-flakes`; the live docs give the machine value as **`nix`**. `nix-flakes` is well-formed but never matches — the same "well-formed ≠ valid" trap recorded in memory. Fixed to `nix`.

Safe-by-construction: a wrong/misspelled ecosystem string degrades to a **miss**, never a false positive — the residual risk is only under-reporting, and the doc URL + fetch date are in a comment above the Set for one-click re-verification.

## Tests

- **Parametrized over all 12** — each must fire. Reverting the Set to `github-actions`-only fails 11 of the 12 rows (falsification-gate verified: exactly 11 fail, the `github-actions` row stays green).
- **False-positive boundary** over 10 semver-supporting ecosystems (`npm, gomod, cargo, pip, uv, bun, maven, gradle, nuget, bundler`) — each asserted **not** flagged.

+22 tests (629 total). `typecheck` clean · `lint` at baseline · `validate_plugins.py` green.

dev-core bumped `0.8.17 → 0.8.18` (semver-disciplined plugin).

### DIFF
--- a/plugins/dev-core/.claude-plugin/plugin.json
+++ b/plugins/dev-core/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "dev-core",
-  "version": "0.8.17",
+  "version": "0.8.18",
   "description": "Full development workflow — frame, shape, plan, implement, review, ship. 30 skills, 9 agents, safety hooks.",
   "author": {
     "name": "Roxabi"
--- a/plugins/dev-core/skills/checkup/__tests__/dependabot-cooldown.test.ts
+++ b/plugins/dev-core/skills/checkup/__tests__/dependabot-cooldown.test.ts
@@ -110,6 +110,50 @@ describe('detectDependabotCooldownViolations', () => {
     )
     expect(detectDependabotCooldownViolations(yml)).toEqual([])
   })
+
+  // The default-days-only ecosystems, per the Dependabot options reference (2026-07-17).
+  // Widening SEMVER_COOLDOWN_UNSUPPORTED beyond github-actions must actually fire for each —
+  // this table is the anchor: reverting the Set to just github-actions fails every row but one.
+  const DEFAULT_DAYS_ONLY = [
+    'bazel',
+    'devcontainers',
+    'docker',
+    'docker-compose',
+    'github-actions',
+    'gitsubmodule',
+    'helm',
+    'nix',
+    'opentofu',
+    'pre-commit',
+    'terraform',
+    'vcpkg',
+  ]
+
+  it.each(DEFAULT_DAYS_ONLY)('flags semver-major-days under %s', (ecosystem) => {
+    const yml = [
+      'updates:',
+      `  - package-ecosystem: ${ecosystem}`,
+      '    cooldown:',
+      '      default-days: 3',
+      '      semver-major-days: 7',
+    ].join('\n')
+    expect(detectDependabotCooldownViolations(yml)).toEqual([{ ecosystem, property: 'semver-major-days' }])
+  })
+
+  // The false-positive boundary: these DO support semver-*-days, so flagging them is the
+  // exact failure this check must never produce. If GitHub's table changes, this fails loudly.
+  const SEMVER_SUPPORTED = ['npm', 'gomod', 'cargo', 'pip', 'uv', 'bun', 'maven', 'gradle', 'nuget', 'bundler']
+
+  it.each(SEMVER_SUPPORTED)('does not flag semver-major-days under %s (semver is valid there)', (ecosystem) => {
+    const yml = [
+      'updates:',
+      `  - package-ecosystem: ${ecosystem}`,
+      '    cooldown:',
+      '      default-days: 3',
+      '      semver-major-days: 7',
+    ].join('\n')
+    expect(detectDependabotCooldownViolations(yml)).toEqual([])
+  })
 })
 
 describe('checkSecurity dependabot.yml status', () => {
--- a/plugins/dev-core/skills/checkup/__tests__/dependabot-cooldown.test.ts
+++ b/plugins/dev-core/skills/checkup/__tests__/dependabot-cooldown.test.ts
@@ -2,7 +2,7 @@ import * as fs from 'node:fs'
 import * as os from 'node:os'
 import * as path from 'node:path'
 import { afterEach, describe, expect, it } from 'vitest'
-import { checkSecurity, detectDependabotCooldownViolations } from '../doctor-local'
+import { checkSecurity, detectDependabotCooldownViolations, SEMVER_COOLDOWN_UNSUPPORTED } from '../doctor-local'
 
 /**
  * Dependabot cooldown validity.
@@ -111,9 +111,34 @@ describe('detectDependabotCooldownViolations', () => {
     expect(detectDependabotCooldownViolations(yml)).toEqual([])
   })
 
+  it('flags a quoted semver key (`"semver-major-days":`)', () => {
+    const yml = [
+      'updates:',
+      '  - package-ecosystem: bazel',
+      '    cooldown:',
+      '      default-days: 3',
+      '      "semver-major-days": 7',
+    ].join('\n')
+    expect(detectDependabotCooldownViolations(yml)).toEqual([{ ecosystem: 'bazel', property: 'semver-major-days' }])
+  })
+
+  it('flags a quoted package-ecosystem value', () => {
+    const yml = [
+      'updates:',
+      '  - package-ecosystem: "docker"',
+      '    cooldown:',
+      '      default-days: 3',
+      '      semver-major-days: 7',
+    ].join('\n')
+    expect(detectDependabotCooldownViolations(yml)).toEqual([{ ecosystem: 'docker', property: 'semver-major-days' }])
+  })
+
   // The default-days-only ecosystems, per the Dependabot options reference (2026-07-17).
   // Widening SEMVER_COOLDOWN_UNSUPPORTED beyond github-actions must actually fire for each —
   // this table is the anchor: reverting the Set to just github-actions fails every row but one.
+  // Independent literal, NOT derived from the production Set — otherwise shrinking the Set
+  // would just drop rows instead of failing them. Drift is instead caught by the equality
+  // assertion below.
   const DEFAULT_DAYS_ONLY = [
     'bazel',
     'devcontainers',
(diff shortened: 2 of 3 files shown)


MickaelV0:
## Code Review — PR #358

Fresh multi-domain review (devops, tester, security-auditor). Every load-bearing claim re-verified by execution (mutation testing) or against the live GitHub docs before posting.

**Verdict: 🔴 Request changes** — the widening is correct and the list is verified complete, but the new per-ecosystem tests have a coverage hole that leaves 11 of the 12 ecosystems' multi-key detection unproven. Since this PR series exists to close exactly that "test passes green without catching the bug" class, it can't ride along.

### What is solid (verified, not asserted)

- **The 12-ecosystem Set is correct AND complete.** devops cross-checked all 12 against the live cooldown support table: the docs list 33 ecosystems, exactly 12 are `default-days`-only (this Set), 21 are semver-capable (the stated complement). No false positive, no omission.
- **`nix` correction is accurate.** GitHub's own Dependabot-Nix changelog gives the machine value verbatim as `nix` (not `nix-flakes`).
- **The `DEFAULT_DAYS_ONLY` anchor is genuine, not hollow.** tester mutation-verified: reverting the Set to `['github-actions']` fails 11/12 rows (only `github-actions` stays green). The false-positive boundary (`SEMVER_SUPPORTED`, incl. npm) already satisfies security-auditor's "assert a semver-supporting ecosystem yields zero findings" ask.
- **No security regression.** All four regexes are linear (no ReDoS), statically constructed (no injection), and the parser's block-scoping is ecosystem-agnostic — widening changes *which* correctly-delimited blocks are inspected, not *how* they're delimited.

### 🔴 Blocker

**1. `issue(blocking)`: per-ecosystem tests only assert `semver-major-days` — a regression dropping `minor`/`patch` detection for the 11 new ecosystems is undetected.**
`plugins/dev-core/skills/checkup/__tests__/dependabot-cooldown.test.ts:117`
Class: `test-tautology`
Mutation-verified by tester: changing `SEMVER_COOLDOWN_KEY` to match only `major` leaves **29/30 green** — the sole catch is the pre-existing `'flags every offending key'` test, hardcoded to `github-actions`. The widened multi-key detection is unverified for the 11 ecosystems this PR adds.
→ **Fix:** cross-product the parametrization over `{major, minor, patch}` (or reuse the multi-key fixture per new ecosystem). — tester (C90)

### 🟡 Warnings

**2. `suggestion(blocking)`: the `DEFAULT_DAYS_ONLY` test literal duplicates the production Set with no drift guard.** — Class: `parallel-path-drift`
`dependabot-cooldown.test.ts:117` ↔ `doctor-local.ts:108`
`SEMVER_COOLDOWN_UNSUPPORTED` is a private non-exported const; the test hand-copies its 12 strings. A future add/remove in prod without mirroring here silently gets zero coverage — the exact drift class this repo has hit before.
→ **Fix:** export `SEMVER_COOLDOWN_UNSUPPORTED` and have the `it.each` iterate the imported Set directly (also resolves #1 cleanly: `Set × {major,minor,patch}`). — tester (C75)

**3. `suggestion(blocking)`: the check silently misses a YAML-quoted key `"semver-major-days":` — a GitHub-valid violation.** — Class: `missing-input-validation`
`doctor-local.ts:110` (`SEMVER_COOLDOWN_KEY`)
Anchored `^semver` means a quoted cooldown key is not matched. Pre-existing (regex unchanged), but this guard's whole value is catching a failure GitHub applies *silently*, so a silent-miss vector is worth closing while we're here.
→ **Fix:** tolerate optional surrounding quotes: `/^["']?(semver-(?:major|minor|patch)-days)["']?\s*:/`, with a test. — security-auditor (C72)

### 🟢 Suggestions (non-blocking)

**4. `nitpick`: comment prose overstates the safety guarantee and misspells one complement value.**
`doctor-local.ts:97,105`
- The "never a false positive" claim holds only for a *misspelled* entry (→ dead miss). A correctly-spelled but *mis-classified* entry (a semver-supporting ecosystem added to the Set) **would** be a false positive — the real dependency this PR rests on. Reword to scope the guarantee to its true mechanism. — security-auditor (C85)
- The complement list writes `rust`; GitHub's machine value is `rust-toolchain` (`cargo` already covers Cargo deps). Doc-only, no logic impact. — devops (C85)

**5. `nitpick`: quoted `package-ecosystem: "docker"` values are untested** (the value regex `['"]?` handles them correctly — verified — but no test locks it). Pre-existing. — tester (C35)

### Praise

`praise:` The falsification discipline held: the anchor table genuinely fails on Set-revert, the false-positive boundary is tested, and the `nix`/`nix-flakes` correction was caught in verification rather than shipped. The findings above are the *next* layer of the same rigor, not a reversal of it.

---

| # | Finding | Class | Blocks | C |
|---|---------|-------|:---:|---|
| 1 | Per-ecosystem tests assert only `semver-major-days` | test-tautology | ✓ | 90 |
| 2 | Test literal duplicates prod Set, no drift guard | parallel-path-drift | — | 75 |
| 3 | Quoted `"semver-major-days":` key silently missed | missing-input-validation | — | 72 |
| 4 | Comment overstates safety + `rust`→`rust-toolchain` | — | — | 85 |
| 5 | Quoted ecosystem value untested | — | — | 35 |

*Reviewed by fresh agents; claims re-verified by execution. Findings only — no code changed.*

MickaelV0:
## Review Fixes Applied — `f4cf918`

All 5 findings from the review above fixed by a fresh fixer (owned both files, file-disjoint from nothing else). **Every fix accepted only after its mutation was confirmed to fail** — the falsification gate was re-run by the orchestrator, not trusted from the agent.

**Applied:** 5 · **Deferred:** 0 · **Skipped:** 0 · **Failed:** 0

### Falsification gate — 4 mutations, all caught (tree restored, 57/57 green at rest)

| Finding | Mutation injected | Result |
|---|---|---|
| 1 — multi-key coverage | drop `minor\|patch` from `SEMVER_COOLDOWN_KEY` | **25 tests fail** |
| 2 — drift guard | revert Set to `['github-actions']` | **36 tests fail** (equality + all rows) |
| 2 — drift guard | add a semver-supporting entry (`npm`) to the Set | **6 tests fail** (equality + npm negative) |
| 3 — quoted key | remove quote-tolerance from the regex | **1 test fails** |

Before these fixes, mutating the key regex to drop `minor|patch` left 29/30 green — the exact hollowness this PR series targets. It now costs 25 failures.

### What changed

**1 (blocking, `test-tautology`):** the positive `it.each` now cross-products `DEFAULT_DAYS_ONLY × {major, minor, patch}` — every semver key is asserted per ecosystem, so the widened multi-key detection is verified for all 12, not just `github-actions`.

**2 (`parallel-path-drift`):** `SEMVER_COOLDOWN_UNSUPPORTED` is now `export`ed. The test keeps its **independent** ecosystem list — deriving it from the Set would destroy the falsification anchor (a Set-shrink would silently drop rows instead of failing them) — and adds `expect(new Set(DEFAULT_DAYS_ONLY)).toEqual(SEMVER_COOLDOWN_UNSUPPORTED)` so drift between the two fails loudly. Independent oracle **and** drift guard, both preserved.

**3 (`missing-input-validation`):** `SEMVER_COOLDOWN_KEY` now tolerates optional surrounding quotes (`/^["']?(semver-…-days)["']?\s*:/`) while still capturing the bare key name, so a YAML-quoted `"semver-major-days":` — valid to GitHub, previously missed — is now caught. Tested.

**4 (nitpick):** the safety comment no longer claims "never a false positive" unconditionally — it now states the guarantee holds only for a *misspelled* entry, while a *mis-classified* (semver-supporting) entry would be a false positive. The complement enumeration was shortened to avoid asserting machine values that rot (the incorrect `rust` was dropped rather than kept).

**5 (nitpick):** added a quoted `package-ecosystem: "docker"` test.

### Gates
`typecheck` clean · `biome` clean on both files · `validate_plugins.py` green · **656 tests** (+27) · `check-skill-version` green (dev-core stays `0.8.18`, already bumped on this branch).

*Fixes by a fresh agent; falsification gate re-run by the orchestrator. `reviewed` label deliberately not set — merge is the author's call.*

MickaelV0:
## Code Review — PR #358 (re-review, iteration 2)

Fresh multi-domain re-review of the fix commit `f4cf918`, targeting whether the 5 applied fixes are genuine and whether they introduced any regression. Every anchor re-verified by independent mutation; every doc claim pulled from the live GitHub source.

**Verdict: ✅ Approve** — all 5 fixes are sound, no new issue, no coverage regression. One non-blocking cosmetic nit, deliberately left as-is (below).

### Verified (mutation / source, not asserted)

- **Every fix anchors exactly its own change** (tester, independent mutations, tree restored each time):
  - drop `minor\|patch` from the key regex → **25 fail** (the cross-product is a real multi-key anchor)
  - add a bogus entry to the Set → **exactly 1 fail**, the equality drift-guard (genuine, not circular — no vacuous pass)
  - revert the Set to `['github-actions']` → **36 fail** (independent `DEFAULT_DAYS_ONLY` literal preserves the original anchor)
  - remove quote-tolerance from each regex separately → **exactly 1 fail each** (the two quoted tests anchor distinct fixes, no cross-masking)
  - **no coverage regression**: the new cross-product asserts the same finding-object shape as the old single-key test, for all three keys — a strict superset (36 vs 12 assertions).
- **No security regression** (security-auditor): the added `["']?` pair is ReDoS-free (optional single-char classes, one greedy `\s*` followed by a literal), the key regex is orthogonal to the block/indent scoping machinery, and exporting the Set widens no surface (plain string Set, sole consumer is the test's drift oracle). Net posture strictly improved — quoted keys (valid to GitHub, previously missed) are now caught, with no false positive traded in.
- **Comment + Set + version correct** (devops, live-table pull): the reworded safety claim matches the code in both directions (misspelled → miss; mis-classified → false positive); `npm/cargo/gomod/pip` are all confirmed semver-capable, correctly excluded; the 12-entry Set is byte-untouched by the fix commit (only `const` → `export const`) and still matches the docs' "Not supported" rows 1:1; `0.8.18` vs staging `0.8.17` is correct (no second bump needed for a same-branch follow-up).

### 🟢 Non-blocking nit (left as-is — intentional)

**`nitpick` (C55, security-auditor): asymmetric-quote matching gives an imprecise *remediation string* on malformed input.** `doctor-local.ts:125`
The two `["']?` are independent, so `'semver-major-days":` etc. match. But every such line is already YAML-invalid / unknown-key → **GitHub rejects the file anyway**, and our check reaches the correct `fail` verdict; only the "keep only default-days there" hint mis-describes the cause on inputs no one hand-writes.
The suggested backreference alternative (`\1`) was **evaluated and rejected**: making the match strict would cause those malformed-but-rejected lines to *stop* matching → the check would report `pass` on a file GitHub rejects → a **new false negative**. The current loose match is the better posture. No change.

### Disposition

Re-review found nothing actionable to fix (the sole nit's correct resolution is leave-as-is). No `/fix` changes applied. The blocking finding from iteration 1 (`test-tautology`) is resolved and mutation-proven. **This is the 2nd review iteration (loop cap reached).**

*Reviewed by fresh agents; claims re-verified by execution and against the live GitHub docs. Findings only — no code changed.*

Issue #353, opened 2026-07-16
fix(promote): rebuild version detection on reachability + stamp version files


MickaelV0:
**Phase 0b of the release-model unification** (analysis: #349 / `artifacts/analyses/release-model-unification-analysis.md`).

**This is the blocking prerequisite.** Nothing else in the migration starts until this ships and is dogfooded. Deleting release-please before this exists freezes the version files in every repo where the bot is their sole writer (`roxabi-intel` `c67362e`/`a067e73`, `roxabi-1page` `17c9394` — all roxabi-ci[bot]) — reproducing the exact roxabi-live `pyproject` bug. **The value of the whole unification is in this fix, not in the deletion.**

## Why — measured, not theorised

`/promote` step 2 detects the base version with `git describe --tags --abbrev=0`, which is **prefix-blind and staging-anchored**. It is **wrong in 7 of 11 repos**:

| repo | step 2 today | correct BASE |
|---|---|---|
| roxabi-plugins | `dev-core/v0.4.4` (another component!) | 0.4.0 |
| roxabi-factory | `artifacts-archive/2026-06` (an archive tag) | (unusable — out of scope) |
| roxabi-talks | **fatal** | 0.2.0 |
| roxabi-forge | **fatal** | 0.4.0 |
| roxabi-live | 0.24.0 (stale) | 0.24.1 |
| voiceCLI | 0.2.1 (stale) | 0.3.0 |
| 1page / intel / llmCLI | ok | ok |

And `/promote` has **no version-file write step at all** — SKILL.md step 2 is `version | V detected`: *detected*, never written. That gap is why `roxabi-live/pyproject.toml` still claims 0.22.3.

## Scope

1. **step 2 — replace `git describe`** with reachability + component scoping:
   ```bash
   COMPONENT=$(yq '.release.component' .claude/stack.yml)
   BASE=$(for t in $(git tag -l "${COMPONENT}/v*"); do
            git merge-base --is-ancestor "$t" origin/main 2>/dev/null && echo "${t#${COMPONENT}/v}"
          done | sort -t. -k1,1n -k2,2n -k3,3n | tail -1)
   COMMITS=$(git rev-list --no-merges origin/main..origin/staging)
   ```
   Component-scoped + main-anchored + tautological payload (= exactly the promotion PR's content).
2. **step 2b — NEW** `write-version-files`: writes `$VERSION` into `release.version_files`. **No-op when the list is empty** (class NONE).
[...]

PR #373, opened 2026-07-19
feat(promote): source-agnostic release via trunk mode (Model B)

### LINKED ISSUE #371: feat(dev-core/promote): source-agnostic release — version from commits, tag+release on merge-to-main (Model B)
## Decision (2026-07-18, Mickael): dev-core `/promote` → source-agnostic ("Model B")

`/promote`'s version **derivation** is already source-agnostic (`price.sh`: `COMPONENT ANCHOR HEAD` → semver from tag reachability + conventional commits; `finalize.ts` takes `isPromote` as a plain bool). But the release **policy** hardcodes the `staging → main` release-train:

- `release-consistency.yml` PR path: `if PR_HEAD_REF != staging → early green` (~1 line)
- `preflight.sh`: `git checkout staging`, `git log main..staging`
- promote `SKILL.md` step 9b: `isPromote` = "newest merged base=main **head=staging** PR"

This staging lock contradicts dev-core's own **project-agnostic** doctrine and blocks making a consuming repo (e.g. roxabi-plugins itself) trunk-based without diverging from the tooling it ships to the fleet.

## Model B (chosen)

Version = pure function of commits since the last `<component>/v*` tag. Tag + GitHub release created **automatically on merge to `main`**. No staging requirement, no pre-declared version, **no three-way witness check** (drop the D7 witness machinery). Effectively `price.sh` run on every merge-to-main = semantic-release with real semver tags.

**Rejected — Model A** (parameterised `release.source_branch`, keep witnesses): more machinery for less gain; still branch-coupled.

### Why
- Removes the staging coupling that contradicts "project-agnostic".
- Unblocks making roxabi-plugins (and any consumer) `main`-only **without** diverging from fleet tooling (factory / live / boilerplate).
- "Versioning at promote" → "versioning at merge-to-main", derived, zero per-PR discipline — but with **real semver tags**, not SHA.

## Scope (spec-level, touches #353)
- Amends `#353` decisions **D8 / D9 / D14** (witness three-way + head=staging policy).
- Config surface: introduce `release.model` (or `release.source_branch`) with a **transition flag** so the fleet migrates incrementally; default preserves current staging behavior until a repo opts in.
- Affects **all dev-core consumers** → needs fleet-wide rollout plan.
- Reconcile the two divergent version surfaces: `plugins/dev-core/.claude-plugin/plugin.json` `version` (marketplace install, bumped by `check-skill-version` pre-push hook — now `0.8.24`) vs `dev-core/v*` git tags (promote).

### Files in blast radius
- `plugins/dev-core/skills/promote/price.sh` — already agnostic, likely reusable as-is
- `plugins/dev-core/skills/promote/lib/finalize.ts` — drop/guard witness path
- `plugins/dev-core/skills/promote/preflight.sh` — remove hardcoded `staging`
- `plugins/dev-core/skills/promote/SKILL.md` — step 9b `isPromote` derivation
- `.github/workflows/release-consistency.yml` — PR-path branch gate
- new: merge-to-main auto-tag/release workflow (or fold into existing)

## Definition of done
- A consuming repo can be configured trunk-based (`main`-only) and still get correct semver tags + releases on merge, using unchanged fleet tooling.
- Existing staging-train repos keep working unchanged until they opt in (transition flag).
- roxabi-plugins can then execute: make `main` default + delete `staging` (currently blocked on this).

## Blocks
- "make `main` default + delete `staging` in roxabi-plugins" — on hold until this ships.

Related: #353 (version detection rebuild), #360 (workflows pure/IO split). Follows PR #370 (48h code-review fixes).

MickaelV0:
## Summary

Makes dev-core `/promote`'s release path **source-agnostic** (Model B). A repo can opt into `release.model: trunk` and get correct `<component>/vX.Y.Z` tags + GitHub Releases created **automatically on every merge to `main`** — no staging→main train, no pre-declared version — using unchanged fleet tooling. `staging-train` stays the default, so the fleet is untouched until it opts in.

The version **derivation** was already source-agnostic (`price.sh` + `finalize.ts`, #353). This wires a thin `release.model` policy layer around it and dogfoods it on roxabi-plugins itself.

- **Config plumbing** — `release.model` flows stack.yml → parser → `WorkflowOpts` → `StackInfo` → drift opts. Absent/unknown ⇒ `staging-train`; only explicit `trunk` activates.
- **`auto-release.sh`** — the sole trunk orchestrator (parentCount → `price.sh` derive → empty short-circuit → `finalize.ts` classify → per-artifact reconcile). **Reuses** price.sh + finalize.ts — no second copy of the 9b logic. Loud-red on stray 1-parent (D3), green no-op on empty payload, per-artifact idempotent recovery (D16).
- **`generateAutoReleaseYml`** — a THIN generated workflow (on:push:[main] + workflow_dispatch, `contents:write`, FIFO `cancel-in-progress:false` + `queue:max`, mint-before-checkout, `fetch-depth:0` + tags) that just invokes the script. Wired into the writer iff `model==trunk` (N18).
- **PR-path gate** — `release-consistency` early-greens every PR under trunk (N7/N8); staging path + push floor unchanged (N9/D14).
- **Double-writer guards** — `/promote`+preflight no-op with `status=trunk_mode` (N17); `/checkup` **fails** on a lingering `release-please.yml` (N10) or an absent/drifted `auto-release.yml` (N11).
- **Dogfood** — roxabi-plugins → `model: trunk`, own `auto-release.yml` generated (byte-equal to generator, N11 verified), release-please trio deleted, all in one changeset.

## Lifecycle

| Phase | Artifact | Status |
|-------|----------|--------|
| Intent | #371: source-agnostic release (Model B) | OPEN |
| Analysis | [371-…-analysis.mdx](artifacts/analyses/371-source-agnostic-promote-analysis.mdx) | Present |
| Spec | [371-…-spec.mdx](artifacts/specs/371-source-agnostic-promote-spec.mdx) | Present |
| Implementation | 7 code commits (S1–S6) + lifecycle artifacts on `feat/371-source-agnostic-promote` | Complete |
| Verification | Lint ✅ · Typecheck ✅ · Tests ✅ (816 pass / 5 skip, ≈51 new) · copy-sync byte-equal ✅ | Passed |

## Test Plan

- [ ] `bun run test` green; `bun run typecheck` + `bun run lint` clean
- [ ] Trunk fixtures: 2-parent→tag, 1-parent→exit1, empty→green-noop, drift→exit1, first→0.1.0, recovery→create-release only
- [ ] `queue: max` schema validity: emitted `auto-release.yml` passes actionlint (only the documented stale-schema false-positive suppressed)
- [ ] Staging-train unchanged: every new guard is inert on the default model
- [ ] Dogfood: `/checkup` reports `release-model:auto-release = pass`, no release-please collision

## SC → Test Matrix

```markdown
| Concern (SC cluster) | Test(s) | Status |
|----|---------|--------|
| release.model parse | parse-stack-yml.test.ts (release block ×6) | ✓ proven |
| opts default + passthrough | workflows.test.ts (normalize + workflowOptsFromStack ×5) | ✓ proven |
| StackInfo release | doctor.test.ts (readStackYml release ×3) | ✓ proven |
| derive→classify→reconcile | auto-release.test.ts (5 topologies + core-reuse + recovery ×11) | ✓ proven |
| emitted-YAML schema validity | auto-release-actionlint.test.ts (×2) | ✓ proven |
| generateAutoReleaseYml W1–W4 + N18 | workflows.test.ts (generator ×5, writer ×2) | ✓ proven |
| PR-path trunk early-green (N7/N8/N9) | release-gate.test.ts (×2 + regression) | ✓ proven |
| preflight trunk no-op (N17) | preflight.test.ts (×2) | ✓ proven |
| N10 collision / N11 missing-drift | workflow-drift.test.ts (×7) | ✓ proven |
| docs 4-concept sentinel (N12/N13) | release-model-docs.test.ts (×6) | ✓ proven |
| N11 dogfood fidelity | live checkWorkflowDrift() | ✓ pass |
```

Every test was confirmed **RED before its implementation** (strict TDD) — the #280 falsification property established inline per pair.

## Scope note

This stages the trunk **config** on roxabi-plugins (release-please → auto-release.yml, both fire on `push:main` — a clean mechanism swap). The **main-default + delete-`staging` branch migration is the separate, now-unblocked follow-up** — this PR does not flip the branch model.

Fixes #371

---


MickaelV0:
## Code Review

**Verdict: 🔴 Request changes** — 5 blocking findings (each independently confirmed by ≥2 fresh domain agents; two empirically reproduced). The feature is well-built and fully green mechanically (816 pass / 5 skip, typecheck ✓, lint ✓, copy-sync byte-equal ✓, `validate_plugins` all-PASS), the no-second-copy invariant and single-writer property hold, and SC coverage is honest — but the dogfood cutover and two silent-failure paths need addressing before this owns roxabi-plugins' releases.

Review: 5 fresh agents (security · architecture · product/spec · tester · devops), single-chunk. Every blocking finding was re-verified against source by the orchestrator.

---

### 🔴 Blockers

**B1 · Trunk dogfood strands roxabi-plugins with no staging→main path** — `product-lead` + `devops` (C90/78)
`.claude/stack.yml:36`
S6 sets `release.model: trunk` on roxabi-plugins while the spec Out-of-Scope **keeps `staging` as default** and the PR-base convention. But N17 (`preflight.sh:24`, `SKILL.md:63`) makes `/promote` **REFUSE/no-op the entire pipeline** under trunk — and `/promote` is the *only* tooled/documented staging→main mechanism (verified: no staging→main workflow exists; default branch is still `staging`). `auto-release.yml` only fires on `push:main`, which nothing now produces. The spec's own safety claim (*"Safe because /promote merges to main still produce 2-parent merge-commits"*) is falsified by the guard this PR ships. A manual `gh pr create --base main --head staging` still works (auto-merges → `push:main` → auto-release fires), but it's undocumented and off-convention.
→ **Fix:** either (a) narrow N17 to guard only the *finalize/tag* step, leaving `/promote`'s create-PR available for repos keeping `staging`; or (b) defer S6 until the main-default migration lands; or (c) ship an interim runbook + have `preflight.sh` print it instead of a bare no-op. Correct the Out-of-Scope safety line either way.

**B2 · `|| true` swallows a `finalize.ts` crash into a false-green `exit 0`** — `security` + `devops` (C88/92, *empirically reproduced*)
`auto-release.sh:100` (falls through `case … *) break` at :120)
`VERDICT=$(bun run …/finalize.ts …) || true` is meant to let the script read the structured `action=refuse` verdict on a non-zero exit — but it also swallows a genuine crash (bun failure, TS throw, arg-parse error). On a crash stdout is empty → `ACTION` empty → wildcard `noop | *) break` → `exit 0`, indistinguishable from a legit empty payload. devops reproduced it: a `finalize.ts` that `throw`s prints the trace to stderr and still exits **0**. Defeats the D3 "loud-red, never a silent release" contract on the write-authoritative path. `Class: bash-error-suppression, missing-error-handling`
→ **Fix:** capture the real exit code (`set +e; VERDICT=$(…); FRC=$?; set -e`) and forgive *only* a well-formed `action=refuse`; assert `[ -n "$ACTION" ]` before the `case`, else `exit 1`.

**B3 · Empty/unvalidated `release.component` silently mis-shifts the generated workflow's args (and is an injection surface)** — `security` + `devops` (C85/75, *empirically reproduced*)
`workflow-generators.ts:224` (dev-init copy `:226`), parser `parse-stack-yml.cjs` bounds component only to `\S+`
`generateAutoReleaseYml` bakes `${component}` **unquoted** into `run: bash …/auto-release.sh ${component} "${{ github.sha }}"`. With `component: ''` the emitted line double-spaces, bash word-splitting drops the empty slot, and `auto-release.sh` receives the **SHA as `$1` (COMPONENT)** with `$2` defaulting to `HEAD` — no refuse (`${1:?}` sees a filled slot), it computes `<sha>/v0.1.0` and would push it. Nothing upstream catches it: D13 component-required never runs (trunk short-circuits), the PR-path early-greens, the push-path skips `read_component` under `version_files: []`, and N11 just byte-diffs generator-vs-generator (both bake empty → "matches"). Separately, `\S+` admits shell metacharacters → latent injection into a `contents: write` workflow across the fleet. `Class: missing-input-validation, shell-injection`
→ **Fix:** throw at generate-time when `component` is empty; pass `COMPONENT` via a quoted `env:` (as `release-consistency.yml` already does for `PR_TITLE`) and reference `"$COMPONENT"`; validate `^[A-Za-z0-9._-]+$`.

**B4 · Trunk PR-path early-green guard is tautologically tested — an inverted guard ships green** — `tester` (C82)
`release-gate.test.ts:151-160` (guard at `release-consistency.yml:128`)
The code is correct, but the test only string-searches: `trunkIdx = reusableSrc.search(/= "trunk" \]/)` matches `= "trunk" ]` **even inside `!= "trunk" ]`**, so flipping `=`→`!=` (inverting the guard, gating trunk PRs and early-greening staging-train) doesn't move the index; the position assertions still hold. Test 1 asserts `read_model()`/`.release.model`/`staging-train` — all present in the *function definition* regardless of the guard. Neither test asserts the branch body is `exit 0`. Violates the repo's #280 falsification norm on the S3 core guard. `Class: test-tautology`
→ **Fix:** extract + spawn the gate (as `auto-release.sh` was, à la `preflight.test.ts`) and assert real exit code with a `model: trunk` fixture; or at minimum anchor the operator (`[^!]= "trunk"`) + assert `exit 0` follows.

**B5 · actionlint schema-validity gate never runs in CI — the safeguard is illusory** — `tester` + `devops` (C80/90, *confirmed*)
`auto-release-actionlint.test.ts:42` + `ci.yml`
The gate is `it.runIf(hasActionlint())`, but `actionlint` is installed by **no** workflow/hook/devDep (verified: referenced only in test + plan files; `ci.yml` has no install step). So in CI it **skips** — only the weak `.toContain('name:' …)` fallback runs, which would pass on a malformed pin or concurrency block. This is the exact "presence ≠ validity / generator-output blind spot" the gate's own docstring says it exists to prevent (dev-core has shipped two nonexistent action SHAs this way). Related: **no CI check byte-compares the committed `auto-release.yml` to the generator** (only `/checkup` N11 does, at human runtime) — a generator edit not mirrored to the committed file passes CI green (`architect`, C88). `Class: generator-drift`
→ **Fix:** add an actionlint install step to `ci.yml` before the vitest step (or fail loud when `process.env.CI && !hasActionlint()`); add a vitest assertion `readFileSync('.github/workflows/auto-release.yml') === generateAutoReleaseYml(roxabiPluginsOpts)`.

---

### 🟡 Warnings (should address; non-blocking)

- **W1 · Generator bakes a roxabi-plugins-internal script path** — `architect` (C82). `workflow-generators.ts:224` hardcodes `run: bash plugins/dev-core/skills/promote/auto-release.sh` — resolves only where dev-core is vendored under `plugins/`. Any *consumer* repo that flips `release.model: trunk` gets a workflow that dies `exit 127` (dev-core lives in `~/.claude/plugins/cache/…`, not the checkout). Spec scopes fleet-distribution out, so it's fine for the dogfood, but N18 emits it for *any* trunk repo with no guard — violates Design Principle #1. → Guard `/ci-setup` to REFUSE trunk when `auto-release.sh` isn't in-repo, + a doc note.
- **W2 · `read_model()` reads PR-head `.claude/stack.yml` to early-green a required gate** — `security` (C68). On `pull_request` the reusable gate checks out the PR merge-ref, so a PR adding `release.model: trunk` flips `release-consistency` to unconditional green. Blast radius is version-consistency (not RCE), diff-visible, and inert on this repo (no ruleset yet) — but def-in-depth erosion for provisioned staging-train repos. → Read the mode from the base ref (`git show origin/main:.claude/stack.yml`).
- **W3 · Real enactment path never tested** — `tester` + `product-lead` (C65/55). Every `auto-release.test.ts` case runs `--dry-run`; `git tag/push` + `gh release create` (`auto-release.sh:111,112,118`) never execute in the suite — a bug there surfaces only on the first real release (the dogfood). → One non-dry-run test with a bare-repo `origin` + a `gh` shim recording argv.

### ❓ Questions (confirm intent)

- **Q1 · `version_files: []` deviates from spec N14** — `architect` (C74). Spec said `[plugins/dev-core/.claude-plugin/plugin.json]`; ship is `[]`. `[]` is arguably *more* correct (plugin.json `0.9.0` is dev-core's **marketplace** line, while `release.component: roxabi-plugins` cuts a *different* tag line) — but it leaves the D14 push-path floor **inert** for the dogfood (git tag = sole version surface). Intended?
- **Q2 · `workflow_dispatch` recovery derives from HEAD, not the failed merge** — `architect` (C66). `SKILL.md:353` names `workflow_dispatch` for partial-failure recovery, but that fires with `github.sha = current main HEAD`; if main advanced, it reconciles a *different* M. The idempotent-recovery guarantee actually holds via GitHub's **"Re-run failed jobs"** (preserves the merge SHA). → Reword the runbook.

### ⚪ Nitpicks

- `auto-release.test.ts:179-211` "no second copy" checks are source-text greps (arg-order not verified) — redundant with the topology tests above them; fine as a lightweight regression guard (`tester`).
- `parse-stack-yml.test.ts:172` "does not confuse a nested component key" has no actual decoy `component:` in the fixture — can't distinguish a scoped parser from a global scan (`tester`).
- SC→Test matrix row "PR-path (N7/N8/**N9**)" bundles N9 (push-path floor) into two PR-path-only tests; holds trivially (push path never reads `release.model`) but the label over-claims (`product-lead`).

### 🟢 Praise

- **App-token trust boundary is correctly constructed** (`security`, C92): `push:main`/`workflow_dispatch` only (never `pull_request_target`), token minted before checkout, least-privilege scope, all actions SHA-pinned, only `github.sha` (quoted) reaches a `run:`, `yaml.safe_load` throughout.
- **No-second-copy invariant + single-writer property are airtight** (`architect`, C97): `auto-release.sh` genuinely reuses `price.sh`+`finalize.ts`; release-please trio fully deleted, atomic delete+add gives exactly one writer with no gap; copy-sync byte-equal (2145/2145).
- **SC coverage is complete and honestly reported** (`product-lead`, C95): all 20 checkboxes trace to real passing tests; the PR body's SC→Test matrix has no inflated "✓ proven" rows.

---

*Generated by `/code-review` — 5 fresh domain agents, findings orchestrator-verified against source. `/code-review` does not fix code; fixing is `/fix`.*

MickaelV0:
## Review Fixes Applied

All 5 blockers resolved across two commits (`6f8144f`, `79c9b1d`); each fix was falsification-verified (broke the target → the guard/test goes red). Full gate green locally: **821 pass / 4 skip**, lint 0, typecheck 0, `validate_plugins` PASS, both workflows actionlint-clean, copy-sync byte-equal.

**Auto-applied (C≥80, 2+ agents):** 4 · **Applied via 1b1:** 1 (B1) · **Deferred (issues created):** 2 · **Skipped:** 3 (praise) + 3 nitpicks

### Blockers — applied
- **[applied] B2** `auto-release.sh` — captured `finalize.ts`'s exit code instead of `|| true`; a crash (empty verdict) is now loud-red `exit 1` (D3), not a false-green. *Falsified: crash → OLD `exit 0` vs NEW `exit 1`; legit `refuse` still `exit 1`.*
- **[applied] B3** `workflow-generators.ts` — validate `release.component` against `^[A-Za-z0-9._-]+$` at generate-time (throw on empty/malformed); pass it via a quoted `env: COMPONENT`. Kills the empty-component SHA arg-shift + the shell-metachar injection surface. Committed `auto-release.yml` regenerated; dev-init copy synced (byte-equal). *Falsified: empty/metachar/space → throws.*
- **[applied] B4** `release-gate.test.ts` — trunk early-green guard test made falsifiable (`[^!]` operator anchor + exit-0 branch probe). *Falsified: inverted `!=` guard → `trunkIdx=-1` → test RED (the old `/= "trunk" \]/` matched inside `!= "trunk" ]`).*
- **[applied] B5** `ci.yml` + `auto-release-actionlint.test.ts` — actionlint now installed in CI (pinned tarball) so the schema gate actually runs; the gate fails loud when `CI && !actionlint`; new vitest gate asserts the committed `auto-release.yml` is byte-equal to the generator (CI analogue of `/checkup` N11). *Falsified: `CI=1` + no actionlint → gate throws.*

### B1 — applied via 1b1 (option a: narrow the guard)
- **[applied] B1** The blanket trunk guard no-op'd the **entire** `/promote` pipeline, stranding a `staging`-keeping repo with no path to `main`. Root cause: it guarded the wrong step. `auto-release.yml` is the sole tagger; `/promote` tags **only** in Step 9 (`--finalize`), never in the create-PR flow (Steps 1–8). Fix:
  - `preflight.sh` — under trunk, **proceed** with the create-PR flow when a `staging` branch exists (`status=trunk_promote_pr`); pure-trunk (no staging) still no-ops (`status=trunk_mode`).
  - `SKILL.md` Step 9.0 — **refuse** `--finalize` under trunk (single-writer guard, now at the tagging step).
  - `SKILL.md` Step 1 table + `## Trunk mode` bullet + **spec Out-of-Scope safety line** corrected — the old *"Safe because /promote merges to main…"* claim (falsified by the blanket guard) holds again under the narrowed guard.
  - Tests: `preflight.test.ts` asserts the `trunk_promote_pr` fall-through (falsifiable vs a revert to the blanket guard); docs sentinel (e).

  **Scope note:** this lifts the *trunk-specific* deadlock. The standard Step 1a gate-probe (release-consistency must be a required check) is unchanged, so roxabi-plugins still needs its `release-consistency` ruleset provisioned before `/promote` runs end-to-end — the same onboarding any repo takes, not a trunk issue. The `main`-default + `staging`-deletion migration remains the separate, now-unblocked follow-up.

### Warnings — dispositions
- **[applied] W3** real enactment path was never tested (all `--dry-run`). Added a non-`--dry-run` `auto-release.test.ts` case: bare `origin` + a `gh` shim → asserts the tag is created at M, pushed, and the release requested exactly once (idempotent, D16).
- **[deferred] W1 → #375** — generator bakes a repo-internal `auto-release.sh` path (breaks *consumer* trunk repos). Confirmed real, but fleet-distribution is out-of-scope for #371.
- **[deferred] W2 → #374** — `release-consistency` reads PR-head `release.model` (a PR can early-green its own gate). Def-in-depth only; inert on this repo (no ruleset yet).

### Skipped
- 3 nitpicks (grep-based checks, decoy fixture, N9 label over-claim) — cosmetic, noted in review.
- 3 praise items — no action.

---
*Generated by `/fix` — fixes applied + falsification-verified; deferred findings tracked in #374/#375.*

MickaelV0:
## Fix follow-up — B1 completeness (adversarial re-review)

A fresh adversarial pass on the B1(a) narrowing (commit `126d4ee`) found the fix was **incomplete**, and **corrects a wrong caveat** in the previous comment:

- **Was broken:** `preflight.sh` opened the create-PR path under trunk (`trunk_promote_pr`), but `SKILL.md` **Step 1a**'s gate-probe still hard-REFUSEd a *protectable* repo with no required, zero-bypass `release-consistency` check on `main` — verified against roxabi-plugins' **live** ruleset. That re-stranded the exact repo B1 unblocks.
- **My earlier caveat was wrong:** I said roxabi-plugins "still needs its `release-consistency` ruleset provisioned." Under trunk that probe should not run at all — there is **no pre-declared version** to guard on a trunk promote PR, and `auto-release.yml` (with its own D3 loud-red guards) is the sole tagger on `push:main`.
- **Fix (`126d4ee`):** Step 1a now short-circuits under `release.model: trunk` — Gate-probe / Unfinalized-promote / Version-file (staging-train *finalize* invariants) are **skipped**; only the **Component** check runs. So the create-PR path is genuinely open end-to-end for a trunk repo that keeps `staging`. Docs sentinel (f) grep-checks the skip; Step 9.0 model read hardened (`-f` guard, fail-safe to staging-train).

**Known/accepted (not blocking, noted for the record):**
- `finalize.ts` is model-blind by design (pure classifier); the `--finalize` trunk refusal lives in Step 9.0's documented guard, enforced like every other skill step. A behavioral test would require threading `--model` through the classifier + both callers — reasonable as a later hardening, out of scope here.
- The `staging`-exists probe is local-ref only (pre-fetch): a main-only/single-branch clone that has never fetched `origin/staging` reads as "no staging" → `trunk_mode` no-op. Low-severity for an interactive `/promote`.

Gate still green: **822 pass / 4 skip**, lint/typecheck/validate clean, both workflows actionlint-clean.

Issue #375, opened 2026-07-20
harden: generateAutoReleaseYml bakes a repo-internal auto-release.sh path (breaks consumer trunk repos)


MickaelV0:
**Origin:** PR #373 review (deferred per `/fix` walkthrough — W1, architect C82, verified REAL by a fresh devops reviewer C90).

`generateAutoReleaseYml` (`plugins/dev-core/skills/shared/workflows/workflow-generators.ts`) bakes a **repo-internal** path into the generated `auto-release.yml`:

```
run: bash plugins/dev-core/skills/promote/auto-release.sh "$COMPONENT" "${{ github.sha }}"
```

This resolves only where dev-core is vendored under `plugins/` (i.e. the roxabi-plugins repo itself). Any **consumer** repo that sets `release.model: trunk` in `.claude/stack.yml` gets a workflow whose `run:` step dies `exit 127` — in a consumer repo dev-core lives in `~/.claude/plugins/cache/…`, not in the checkout. The N18 writer (`workflow-push.ts`) emits it for **any** trunk repo with no path-existence guard (violates dev-core Design Principle #1 "project-agnostic").

In scope for #371 this is acceptable — the spec scopes fleet distribution out and only roxabi-plugins dogfoods trunk. This issue tracks the fleet-distribution hardening.

**Action:** guard `/ci-setup` (or the N18 writer) to REFUSE trunk emission when `auto-release.sh` is not resolvable in-repo, and/or ship the script to consumer repos + reference a portable path. Add a doc note.

Issue #374, opened 2026-07-20
security: release-consistency reads PR-head release.model — a PR can early-green its own gate


MickaelV0:
**Origin:** PR #373 review (deferred per `/fix` walkthrough — W2, security C68).

`release-consistency.yml`'s `read_model()` reads `release.model` from the **PR-head** `.claude/stack.yml` (the reusable gate checks out the PR merge-ref). A PR that adds `release.model: trunk` therefore flips `release-consistency` to unconditional early-green **for its own PR** — a defence-in-depth erosion for provisioned staging-train repos.

Blast radius is version-consistency only (not RCE), the change is diff-visible, and it is inert on roxabi-plugins today (no `release-consistency` ruleset provisioned yet). Hence deferred, not blocking.

**Action:** read the release mode from the **base** ref instead of the PR head, e.g. `git show origin/main:.claude/stack.yml` (mirror `read_component` if it has the same issue).

PR #403, opened 2026-08-02
fix(dev-core): classify analyses by frontmatter, not filename

MickaelV0:
## Summary

Fixes the `/dev` infinite loop when `artifacts/analyses/` holds more than one kind of document (legacy consensus, brainstorms, analyses). Replaces hand-rolled bash/awk classification with a tested TypeScript classifier, hardens every writer to the same frontmatter contract, and de-fangs issue titles that reached `git commit -m` / YAML.

### 1. Shared-directory α resolution

`artifacts/analyses/` historically held analyses, brainstorms, and consensus artifacts. Resolvers used alphabetical `head -1` → a consensus file sorted ahead of the real analysis and wedged `/dev` forever (`Σ.analyze` permanently false).

**Fix trajectory on this PR:**
- Filename filter → rejected in review (wrong axis)
- Frontmatter sort via bash/awk → three review rounds of fence bugs (CRLF/BOM/space, unterminated fence, first-wins vs YAML last-wins)
- **Final:** `artifact-classify.ts` (bun) — fence-correct, fail-closed on unterminated, YAML last-wins; `lib.sh` is a thin wrapper

Also: `/consensus` removed; brainstorms write to `artifacts/brainstorms/`; `/pr` `ISSUE_NUM` accepts any conventional prefix (`fix/`, `chore/`, …), not `feat/`-only.

### 2. Title injection + writer contract

Issue titles are external content. They reached commit subjects (`$(…)` execution) and unescaped YAML (`status:` injection). **Title hygiene** + required `type:`/`status:` on write now live in one SSoT (`skills/shared/references/artifact-frontmatter.md`), applied by `/analyze`, `/interview`, `/spec`, `/frame`, and review skills.

### Test plan

- [x] Mixed dir (consensus + analysis) → resolve picks analysis
- [x] Consensus-only → empty resolve (recovery path)
- [x] CRLF / BOM / trailing-space fences → correct status
- [x] Unterminated fence → `malformed`, not body-as-header
- [x] Duplicate `status:` → last-wins
- [x] `type: analysis` beats `status: consensus-reached` (falsifying test)
- [x] `fix/42-…` branch → `/pr` reports `issue=42`
- [ ] CI green on this push

## Version price

`price.sh roxabi-plugins` → **0.12.0** (minor). No `!` / BREAKING CHANGE footer (by design: marketplace is one component).

---
Generated with [Roxabi dev-core](https://github.com/Roxabi/roxabi-plugins) via `/ship`


MickaelV0:
## Code Review

15 findings. The orchestrator re-verified the load-bearing ones against source before recording them; results below distinguish confirmed from corrected.

**Verdict: do not merge. The approach is wrong, not just incomplete.**

### The headline — the recovery path this PR claims does not exist

`analyze/SKILL.md:101` globs `artifacts/analyses/*` **unfiltered**. So the fix moves the deadlock one step later rather than removing it:

`scan-state` now reports `analyze=false` → `/dev` dispatches `/analyze` → `/analyze` re-globs unfiltered, re-picks the consensus artifact, reads `status: consensus-reached`, and its Step 1 table has **no row** for that token — the exact undefined behaviour this PR's description diagnoses. Worse, if the model takes the "load as base" row and the user approves, `status: approved` lands on a `-consensus.md` file that the new filter excludes → `analyze=false` again → the same infinite loop.

The PR's own test-plan item — *"consensus-only directory → `analyze=false`, `/dev` runs `/analyze` and can complete"* — passes at the scan-state layer and still deadlocks one layer down. A green checkbox on a broken chain.

### Root cause: filtering on filename is the wrong axis

The authoritative kind marker is **frontmatter**, not the filename. `/analyze`'s own Step 1 table proves it — it has a `type: brainstorm` row. Keying the resolver on a filename substring was chosen because it is what `scan-state.sh` can see cheaply, and every finding below is a consequence.

**False negatives** — real approved analyses become invisible, flipping `Σ.analyze` true→false (verified by running the pattern):

| filename | matches `-analysis\.md` |
|---|---|
| `42-x-analysis.md` | ✅ |
| `42-x-analysis.mdx` | ✅ |
| `42-x-analysis-iter2.mdx` | ❌ — the `-iterN` convention already in this directory |
| `42-x-analysis.claude.md` | ❌ — the fleet suffix in `ssot/conventions.ssot.md` |
| `42-x.md` | ❌ — `interview/SKILL.md:51-52` documents α and β as `artifacts/analyses/{slug}.md` |

(`interview` is itself inconsistent: `:51-52` say `{slug}.md`, `:121` says `{slug}-analysis.md`.)

**False positive** — `/interview` writes brainstorms to `{slug}-analysis.md` with frontmatter `title`/`description`/`type: brainstorm` and **no `status` key** (`references/templates.md:7-14`). It passes the new filter, and `dev/SKILL.md:87` treats missing status as approved → `Σ.analyze` true → `/dev` skips analyze entirely and drives `/spec` off a brainstorm.

**Unanchored** — `42-x-analysis.md.orig` matches (verified) and sorts before `.mdx` under `head -1`, so a rebase leftover becomes the gate signal. Also `\-` is undefined in POSIX ERE; `-analysis\.mdx?$` is correct and portable.

### Two defects this PR introduces

- **`git add artifacts/reviews/{file}`** — `{file}` is defined nowhere, and the hardcoded prefix contradicts `:197` (`N missing → {slug}-adversarial.md`, a bare filename). The user consents, `git add` fails on a non-existent path, `&&` short-circuits, no commit — surfacing as a raw git error. Same line in `advisory:215`.
- **`{slug}` as commit subject** — `:195` documents the "slug unsafe" branch. A title with no ASCII alphanumerics yields an empty slug → `git commit -m "docs(adversarial): "` → commitlint `subject-empty` rejects it (wired in this repo via `lefthook.yml`). Injection bug traded for a hard commit failure.

### Incomplete coverage

- **`/consensus` has both identical title sinks** — `:156` `title: "{title} — Expert Consensus"` and `:209` `docs(consensus): {title}`. This PR *newly teaches `/spec` to trust κ's `status:` as a gate signal* while leaving κ's writer unhardened — it raises the value of injecting there. Same pair at `analyze:144`, `plan:168`, `spec/references/templates.md:11`, `frame:178` (unquoted).
- **The glob collision remains in five more resolvers** — `adversarial:66`, `advisory:69`, `pr/gather-state.sh:31`, `clarify:77`, `plan/references/plan-template.md:30`. Note `plan/references/micro-tasks.md:120` already uses `{issue}-*-analysis.md`, so the filtered form was the established convention this PR only partially restored.
- **`/spec` priority order** — glob #1 (α) pre-empts #2 (κ), so a stale *draft* analysis beats the consensus written specifically to supersede it: STOP "Analysis is still draft", consensus never consulted.
- **κ half-introduced** — `:22` still reads `SRC := source doc (α ∨ φ)`, the Let block never defines κ, the no-source STOP text never mentions `/consensus`, and `spec/README.md:24,41` are stale.
- **No regression test** — `scan-state.sh` feeds every gate decision in `/dev` and has no test anywhere, in a repo with vitest and 20+ test files. These are pure shell assertions over a fixture directory.

### Corrected from the finder's report

`-analysis.mdx` **does** match the pattern (substring, not anchored), so the `.mdx` half of that finding does not hold. The `-iterN` and `.claude.md` halves do.

### Direction

Decide kind on frontmatter, not filename: `scan-state.sh` emits candidates, the caller reads `type:`/`status:` to classify. That closes the false negatives, the brainstorm false positive, and the `.orig` case together, instead of tuning a regex against a naming convention that has at least four live forms.

MickaelV0:
## Code Review

4 agents (architect · tester · security-auditor · devops), 24 files / 998 lines, single chunk. Secret scan clean, no spec artifact → product-lead skipped. Every headline claim below was re-verified by the orchestrator against source; commands and outputs are quoted.

**Verdict: Request changes.** One finding fires irreversibly on merge and is not a code defect.

### ⚠️ Merging cuts `roxabi-plugins/v1.0.0`

```
$ bash plugins/dev-core/skills/promote/price.sh roxabi-plugins roxabi-plugins/v0.11.0 HEAD
1.0.0
```

`price.sh` derives **one** version for the whole marketplace from every commit since the last tag, and treats any `!` / `BREAKING CHANGE:` as an unconditional major bump — no per-plugin scoping, no pre-1.0 leniency. `eb0a0ce feat(dev-core)!: remove /consensus` is **the first `!`-marked commit in this repo's history**, so this path has never fired. Meanwhile `plugin.json` was hand-bumped 0.10.9 → 0.11.0, the conventional pre-1.0 reading. The two schemes disagree and only the repo-wide one runs in CI, unreviewed, on push to `main`.

Fix before merge: drop the `!` and the footer (the plugin.json minor bump already records the break at plugin scope) — or state explicitly that a v1.0.0 marketplace release is intended.

### Blockers — the classifier

**The frontmatter window is defined by line count, not by the `---` fence.** `head -30` reads the first 30 lines of the *file*, so body text is indistinguishable from a declaration. Verified, both directions:

```
# an approved analysis that merely QUOTES the marker in a fenced block
$ scan-state.sh --classify-artifact 42-auth-analysis.md   → consensus|approved
$ scan-state.sh --resolve-analysis . 42 auth              → ""
```

That is **worse than the bug being fixed**: the original wedge was recoverable, this one is not — re-running `/analyze` rewrites the same file, keeping the fence, so the analysis stays invisible forever. And it is highly reachable here: the most likely document to quote `status: consensus-reached` is an analysis of this very defect. `interview/references/templates.md:13` already carries `^type: brainstorm` at line start inside a fence.

Fail-open in the other direction: 31-line frontmatter → marker invisible → `analysis|` → key-absent ≡ approved → a draft clears the Shape gate. Same window bug in `artifact_status`, so a body line `status: draft` can outrank the real key.

**Fix:** scan between the first `---` and the next, and drop the line cap. One edit closes all three.

**The classifier fails open on missing frontmatter.** `374-375-hardening-spec.claude.md` opens with an HTML comment — no frontmatter at all → `analysis|`. Had it stayed in `artifacts/analyses/` it would have resolved as an **approved α** for #374 and #375. The manual move (subject 3) is the only thing protecting against the class subject 1 claims to solve. Root cause: the discriminator is a *denylist of two known-bad kinds*, not an allowlist — there is no positive `type: analysis` marker to test. `.orig` / `.rej` leftovers classify as analyses too.

### Blockers — coupling between the three subjects

- **`spec/SKILL.md:75` still carries the pre-removal ordering** — `Prefer an approved α; else κ; else a draft α` — 21 lines above `:96` which says a consensus artifact is not a valid SRC and STOPs. Subject 2 rewrote the outcome rules without revisiting the ordering rule they were derived from.
- **`/interview` lost its own read path.** Step 1 is `Glob A, S` — B is absent. Before the move it found brainstorms *because* they lived in A; now a second `/interview` on the same topic cannot see the brainstorm it wrote itself, so the promote/build-on gate never fires. Same gap in `clarify/SKILL.md:77`. (`/analyze` and `/spec` were correctly updated; `/frame`, `/plan`, `/pr` legitimately don't consume β.)
- **The convention ships with zero instances.** `brainstorms` was added to this repo's `stack.yml` only. Verified missing from `.claude/stack.yml.example`, `plugins/dev-core/stack.yml.example`, `stack-setup/SKILL.md` — plus `dev-init/lib/scaffold.ts` (mkdir list), `scaffold-rules.ts` (defaults + the generated CLAUDE.md Artifact Model table). A freshly initialized repo gets a writer pointing at a directory init never creates.

### Blockers — parallel paths

- **`gather-state.sh` copies the rule but not the candidate set.** `scan-state.sh` uses anchored `(^|[^0-9])N-` + slug fallback; `gather-state.sh` uses glob `N-*` with no fallback, while its comment claims "same rule". Three verified divergences where `/dev` reports an analysis and `/pr` reports `analysis=false` — including `dark-mode.md`, a form `resolve_analysis` supports and its own test pins. Both scripts already source the same `lib.sh`; the helpers belong there.
- **An issue title still injects `status: approved`.** The title-hygiene rule was scoped to the two skills where the sink was introduced, not to the sink class. `/analyze:149` renders `title: "{title}"` directly above `status: draft`, and `artifact_status` takes `head -1` of the first match — an injected key wins on line order. Verified end-to-end: `analysis|approved` on a draft. `/frame:178` is worse (unquoted). This PR raises the payoff by making that field a machine-read key.
- **`status` value is not normalized.** `status: "approved"`, `status: approved # was draft`, `Status: Approved`, block scalars — all valid YAML meaning approved, all stall `/dev` permanently.

### Tests

The falsification pass was run properly: main's pre-fix resolver was ported back under the new hooks. **3 of the 7 `resolve_analysis` tests fail (correct), 3 pass unchanged.** Those 3 pin pre-existing anchor/slug behaviour this PR didn't touch, but sit under a `describe` labelled "the alphabetical-pick regression" — a future simplification dropping the kind check leaves them green. Regroup them.

Also: no fixture models the no-frontmatter / HTML-comment shapes that exist in this repo today (3 real files); the live dir-iteration loop around `resolve_analysis` (new code in this PR, sets `ANALYZE_DIR`) is uncovered; `expect(r.status).toBe(0)` swallows stderr, so a portability break fails opaquely across all 11.

### Cleared

- **`/consensus` removal is clean** — no dangling `Skill()` call, marketplace entry, or chain step. Redirect rows and successor lists updated; legacy branches STOP with a pointer to a live skill.
- **Copy-sync green** — `validate_plugins.py` 12/12, `sync-shared --check` clean, `@generated` header intact, 852 tests pass.
- **Filename handling is injection-safe** — `$(…)`, backticks, `;`, `*`, spaces all resolve correctly with zero execution. `IFS= read -r`, `grep -F --`, heredoc-not-pipe all verified sound.
- **Frontmatter-over-filename was the right axis** — on the real 12-artifact corpus it separates all 4 consensus files (including `pr184-…`, which no `{N}-` anchor would catch) from the 8 analyses, no false positives.
- Disproved by the auditor and not reported: `set -euo pipefail` aborting on grep no-match; SIGPIPE+pipefail inverting the skip; `for f in glob` leaking the pattern; untrusted args reaching a command via the new hooks.

### Recommendation

`analyze_status` is emitted but **nothing reads it**, and `dev/SKILL.md:84` still says the opposite ("emits filenames only… ¬by the helper's truthy line") — `dev/SKILL.md` took only a version bump. Either wire it as the single parser or drop it.

Separately: `ports/artifacts.ts` has **0 importers** (`ports/config` has 5, `ports/workspace` 7). ADR-015 kept it by name on a claim that does not hold; this PR spent a copy-sync round-trip on a type no code reads. Delete it and amend the ADR — as a follow-up, not here.

**On the shape of the PR:** subjects 1 and 3 are competing solutions to one problem — segregation makes classification unnecessary for new writes, classification exists only for files already on disk. Shipping both is defensible as target-state + migration, but this PR ships the *write* side of the target state, the *read* side of the migration, and omits the scaffolding side entirely. Subject 2 is independent and breaking. 2 and 3 each warranted their own PR.

MickaelV0:
## Code Review — round 4

4 agents on the delta `613d374..edfb8af`. **Verdict: Request changes.** The round-3 fix closed the defect it targeted and introduced two new ones in the same function. Architect and security-auditor found both independently; re-verified by the orchestrator.

### The awk replacing `head -30` is wrong in two ways

```
$ scan-state.sh --classify-artifact <fixture>
crlf.md          (---\r\n … status: draft)          → analysis|      ← should be draft
trailws.md       ("--- " … status: consensus-reached) → analysis|      ← should be consensus
unterminated.md  (--- … no closing fence, body has   → consensus|approved
                  status: consensus-reached)
```

**1. The open fence uses `==`, the close uses a regex.** `NR==1 && $0!="---"` rejects `---\r`, `--- `, or a BOM, so the scan aborts at line 1 and returns *no frontmatter* — which `/dev` reads as key-absent ≡ **approved**. This is a **regression**: the `head -30` grep it replaced anchored on `[[:space:]]*$`, and `[[:space:]]` matches `\r`, so the old code got these right. dev-core ships to repos whose line endings it does not control; `.gitattributes` here pins only `*.ts`.

**2. An unterminated fence prints the whole file.** The awk streams `{print}` with no "was a close seen?" guard, so on `---` + no closing fence the window becomes the entire document — *wider* than the `head -30` it replaced, and it restores exactly the body-vs-declaration confusion this round claimed to remove. Unrecoverable: a genuine analysis classified `consensus` is skipped forever, and re-running `/analyze` rewrites the same path.

Both fixes are one edit: strip `\r`, use the same regex for both fences, and buffer — emitting only on a confirmed close.

### The `/pr` ↔ `/dev` divergence was not closed

```
gather-state.sh:28  ISSUE_NUM=… sed -n 's/^feat\/\([0-9]*\)-.*/\1/p'   ← feat/ only
gather-state.sh:35  SLUG=…      sed -n 's/^[a-z]*\/[0-9]*-\(.*\)$/\1/p' ← any prefix
```

The candidate set was unified; the input derivation was not. `ISSUE_NUM` gates the whole block, so on `fix/{N}-` and `chore/{N}-` branches `/pr` still reports `issue=none` — no analysis, no spec, no test count. Four of the last ~25 merged PRs here used those prefixes. `[a-z]*` in the new sed is unreachable generality.

### Hardening landed on one α writer of two

`/interview` also writes α into `artifacts/analyses/` (`templates.md:41`), and its template emits `title` + `description` only — **no `type:`, no `status:`, unescaped `{title}`**. So an `/interview`-authored analysis is born **approved** (absent key ≡ approved), bypassing the HITL gate `/analyze` was given, and its title is an unguarded sink into the frontmatter the new classifier parses. Same omission in `/spec`'s σ template (`spec/references/templates.md:11`), whose `status:` **is** a gate (`dev/SKILL.md:97`) — verified: an injected key lands above the real one and `artifact_status` takes the first match.

Related: `/interview`'s β→α promote has no instruction to drop `type: brainstorm`, so a promoted analysis can carry the key that makes it invisible.

### Sink semantics

`artifact_status` uses `head -1` (first match) where **YAML resolves duplicate keys to the last**. The `/analyze` hygiene note names this exact hazard and hardens the source; the sink keeps the divergent rule, so every unhardened writer inherits it. Also `status:`, `status: ` and `status: ""` all normalize to empty → approved: a present-but-empty gate value fails open.

### Tests

- **`type: analysis` has no falsifying test.** The positive branch and the `else` default both return `analysis`, and no fixture writes `type: analysis`. It only diverges when combined with a competing key — the precedence the docstring promises is untested.
- `resolve('', 'quoting')` passes an empty N, so it exercises the slug fallback, not the N-anchored path real callers use.
- The two no-frontmatter fixtures hit the identical branch (neither contains `---`); one is redundant.
- `pr/gather-state.sh` has **no test file at all**, and this PR gave it a new sed and a new call.
- None of the four shapes above (CRLF, unterminated, empty value, duplicate key) is covered.

### Cleared

- **The reword is clean.** `git rev-parse 2dad3c7^{tree} edfb8af^{tree}` → identical hash; messages only. `price.sh` → `0.12.0`; zero breaking markers in the payload; commitlint passes on all 6 commits. Release notes derive from **PR titles**, not commit bodies, so no other consumer is affected.
- **Sourcing contract holds** — zero variables leaked, zero shell-option delta, zero commands at source time; `awk` is confined to one function body.
- **Status normalization is right**, including `status: approved#nospace` correctly *not* stripped (YAML requires whitespace before `#`).
- **Injection-safety re-verified** at the new callsite: `$(…)`, backticks, `;`, `*`, spaces, `-rf` all inert.
- All 12 real artifacts still classify correctly; 12/12 validation, 857 tests, copy-sync clean, typecheck clean.
- Disproved and not filed: `---` inside a block scalar truncating frontmatter; 10MB-line or `---`-only files breaking the parser; option-injection via `grep -F`; `/adversarial` and `/advisory` being unhardened.

### Also

7th scaffolding site missed: `plugins/dev-core/references/dev-process.md:36` carries the same Artifact Model table and still shows four rows, while `scaffold-rules.ts` now generates five for user repos. `/plan` reads that file directly. And `shared/__tests__/fixtures/sample-stack.yml` is drift-only (no assertion on `artifacts`).

### Note

An agent observed `lib.sh` being mutated in the working tree mid-review by something outside the session, then reverted (`grep -E "$anchor"` → `grep -E "THIS_IS_BROKEN_NEVER_MATCHES"`). Results above were re-run against the pinned HEAD blob with sha256 confirmed. Working tree is clean at `edfb8af`. Flagging because it also explains one transient test failure in that agent's first run — not a PR defect.

MickaelV0:
## Code Review

**Scope:** `origin/main...HEAD` on #403 (39 files) — classifier rewrite, frontmatter contract, writer hygiene, brainstorms dir, `/consensus` removal.
**Agents (orchestrator + probes):** security, adversarial, tester, architecture. No FE/BE app code. No axial ADR paths.

### Secret scan
∅ matches.

### Spec compliance
No issue-linked spec for this branch (hotfix-style pipeline fix). SC→Test matrix N/A (τ S / no product SCs).

---

### Findings

#### Praise

praise: Classification finally leaves the bash/awk failure class. Fence extraction + last-wins keys + fail-closed unterminated are unit-tested; `lib.sh` is a pure dispatch wrapper so the two resolvers cannot drift again.
  `plugins/dev-core/skills/shared/artifact-classify.ts`
  -- architect
  Confidence: 95%

praise: Shared `artifact-frontmatter.md` closes the write-side hole that only hardened `/analyze` — `/interview` and `/spec` now emit `type:` + `status: draft` + yaml-escaped titles; sub-specs and promote detection follow.
  `plugins/dev-core/skills/shared/references/artifact-frontmatter.md`
  -- product-lead
  Confidence: 90%

praise: Regression suite covers the tour-4 blockers (CRLF/BOM/space, unterminated → malformed, last-wins, `type: analysis` falsifier, shell integration hooks).
  `plugins/dev-core/skills/shared/__tests__/artifact-classify.test.ts`
  -- tester
  Confidence: 95%

#### Suggestions (non-blocking)

suggestion(non-blocking): `bun` is now a hard runtime dependency of `/dev` state scan. Missing bun → exit 127 under `set -e` (probed). Acceptable for this fleet (bun already required), but a one-line fallback message in `scan-state.sh` when `artifact_classify` fails would turn a cryptic empty Σ into an actionable STOP.
  `plugins/dev-core/skills/shared/lib.sh:168`
  -- devops
  Root cause: wrappers return 127 without a scan-state-level recovery path
  Solutions:
    1. On non-zero classify, emit `analyze=false` + `analyze_error=bun-missing` and continue (recommended if we want soft degrade)
    2. Keep hard fail; document bun as scan-state prerequisite (status quo)
  Confidence: 70%

suggestion(non-blocking): `ISSUE_NUM` regex is lowercase-only (`^[a-z][a-z0-9]*\/`). `Feat/42-…` still yields `issue=none`. Conventional branches are lowercased in this fleet — fine; note only if case-preserving remotes appear.
  `plugins/dev-core/skills/pr/gather-state.sh:28`
  -- backend-dev
  Root cause: sed class omits A-Z
  Solutions:
    1. Case-fold branch before sed
    2. Leave as-is (recommended for now)
  Confidence: 60%

nitpick: Empty frontmatter fences (`---\n---`) still default to `analysis|` (legacy approved). Documented and intentional for three legacy files; empty-but-fenced is rare.
  `plugins/dev-core/skills/shared/artifact-classify.ts:80`
  -- adversarial
  Confidence: 55%

---

### Adversarial probes (manual)

| Case | Result |
|------|--------|
| CRLF / BOM / `--- ` fences | `analysis\|draft` |
| Unterminated + body `status: approved` | `malformed\|` (not approved) |
| Duplicate status keys | last-wins (`approved` after `draft` → approved) |
| Mixed consensus + analysis dir | resolve picks analysis |
| `bun` not on PATH | exit 127 + stderr |

No residual bash/awk frontmatter parsers under `skills/`.

### Verdict

**Approve** — no blockers. Non-blocking suggestions only; fix loop not required.

`price.sh roxabi-plugins` → **0.12.0** (minor). Safe to gate with `reviewed` + auto-merge once CI is green.

